PR #1, opened 2016-10-03
Upload stemcell

zachgersh:
would love feedback on this - currently missing the progress bar but that will be added shortly.

- authenticated client makes an appearance
- added counterfeiter fakes for consistency, will update the rest of the fakes in another PR
- add dependencies for oauth 2 (ignore this commit on review as there are so many files)
- renamed OpsManager to Ops Manager

### TODO:

- [X] add progress bar code back in
- [X] refactor token acquisition


zachgersh:
@davewalter this is to make the api "symmetrical". Want to make sure that even if an empty response is returned that we still have that struct around. For few things we implement there should be a very consistent pattern to follow.

zachgersh:
@aditya87 great idea on the shared server setup for network - that is fixed. I don't believe I understand your other question. The test that covers 46 is "when the origin file cannot be read". You basically cannot make it fail at that point if the file will not open.

PR #3, opened 2016-10-12
Added export installation command with progress bar

aditya87:
- Targets an ops manager and exports the full installation to a local
  file by getting the "installation_asset_collection" API endpoint

See tracker story: https://www.pivotaltracker.com/story/show/132128225

Signed-off-by: Aditya Anchuri <aanchuri@pivotal.io>


acceptance/export_installation_test.go
```diff
@@ -0,0 +1,122 @@
...
+	Context("when an error occurs", func() {
+		Context("when the output file cannot be written to", func() {
+			var badOutputFile *os.File
+
+			BeforeEach(func() {
+				var err error
+				badOutputFile, err = ioutil.TempFile("", "")
+				Expect(err).NotTo(HaveOccurred())
+
+				err = os.Chmod(badOutputFile.Name(), 0000)
```
zachgersh: can we not rely on permissions? lets just remove the file, gives you a similar failure mode but we don't have to run as a different user.

acceptance/help_test.go
```diff
@@ -22,6 +22,7 @@ Usage: om [options] <command> [<args>]
 
 Commands:
   configure-authentication  configures Ops Manager with an internal userstore and admin user account
+  export-installation       exports the installation of the target ops manager
```
zachgersh: please keep all of the mentions of `Ops Manager` consistent.

api/installation_service.go
```diff
@@ -0,0 +1,53 @@
...
+		return fmt.Errorf("request failed: unexpected response")
+	}
+
+	is.progress.SetTotal(resp.ContentLength)
+	is.progress.Kickoff()
+	progressReader := is.progress.NewBarReader(resp.Body)
+
+	respBody, err := ioutil.ReadAll(progressReader)
+	if err != nil {
+		return fmt.Errorf("request failed: response cannot be read") //Can't test drive
```
zachgersh: You can remove the comment, it is fine that it is not covered by a test 💃

api/installation_service.go
```diff
@@ -0,0 +1,53 @@
...
+
+	if resp.StatusCode != http.StatusOK {
+		return fmt.Errorf("request failed: unexpected response")
+	}
+
+	is.progress.SetTotal(resp.ContentLength)
+	is.progress.Kickoff()
+	progressReader := is.progress.NewBarReader(resp.Body)
+
+	respBody, err := ioutil.ReadAll(progressReader)
```
zachgersh: we should be closing the response body. I screwed this up as well in my service :/

api/installation_service_test.go
```diff
@@ -0,0 +1,124 @@
...
+			Expect(err).NotTo(HaveOccurred())
+
+			By("posting to the installation_asset_collection endpoint")
+			request := client.DoArgsForCall(0)
+			Expect(request.Method).To(Equal("GET"))
+			Expect(request.URL.Path).To(Equal("/api/v0/installation_asset_collection"))
+
+			By("writing the installation to a local file")
+			ins, err := ioutil.ReadFile(outputFile.Name())
+			Expect(err).ToNot(HaveOccurred())
```
zachgersh: flip this back to `NotTo`

api/installation_service_test.go
```diff
@@ -0,0 +1,124 @@
...
+			var err error
+			client = &fakes.HttpClient{}
+			outputFile, err = ioutil.TempFile("", "")
+			Expect(err).ToNot(HaveOccurred())
+			bar = &fakes.Progress{}
+		})
+
+		AfterEach(func() {
+			err := os.Remove(outputFile.Name())
+			Expect(err).ToNot(HaveOccurred())
```
zachgersh: also `NotTo`

api/installation_service_test.go
```diff
@@ -0,0 +1,124 @@
...
+			client     *fakes.HttpClient
+			outputFile *os.File
+			bar        *fakes.Progress
+		)
+
+		BeforeEach(func() {
+			var err error
+			client = &fakes.HttpClient{}
+			outputFile, err = ioutil.TempFile("", "")
+			Expect(err).ToNot(HaveOccurred())
```
zachgersh: same.

api/installation_service.go
```diff
@@ -0,0 +1,53 @@
...
+}
+
+func NewInstallationService(client httpClient, progress progress) InstallationService {
+	return InstallationService{
+		client:   client,
+		progress: progress,
+	}
+}
+
+func (is InstallationService) Export(outputFile string) error {
```
zachgersh: There was some discussion around the pattern for each one of these services today. The currently established pattern is to take in a struct for the input and return a struct + error for the output (even if ignored).

I see a couple of benefits to this but I welcome being challenged on this:
- Services have symmetry (all of them take an input and produce an output) this makes the API "guessable"
- If we ever decide to add / remove input or output arguments we end up only needing to touch the struct (and the implementation feels future proof without adding a higher support cost).

Does that make sense?

aditya87: I personally don't see the gain in returning a struct if it's always going to be empty, simply to follow an "established pattern". I see your point about wanting to add/remove input/output arguments later, but I think if we ever need to add them then we can create a struct then. I think there's a balance to be struck between making something future proof and anticipating functionality that may not be there. 

Different services can have different interfaces, I guess that's the nature of having different commands/tools that do different things. If every interface looked the same it wouldn't serve much of a purpose besides simple. symmetry.

acceptance/export_installation_test.go
aditya87: Also we can't remove the file, since the command will create a file if it doesn't exist :(

api/installation_service.go
aditya87: Done

api/installation_service_test.go
aditya87: Done

acceptance/export_installation_test.go
zachgersh: @aditya87 what happens if you give it a directory that doesn't exist at all? it won't recursively create the structure - right? So can't we do something like "/unknown/my-temp-out"

aditya87: Hmm yeah guess we could do that.

aditya87: Done.

PR #5, opened 2016-10-14
Import installation

aditya87:
Command to import installation into a deployed, unconfigured ops manager from a local file.

Usage:
```
ॐ  import-installation
This unauthenticated command attempts to import an installation to the Ops Manager targeted.

Usage: om [options] import-installation [<args>]
  -v, --version              bool    prints the om release version (default: false)
  -h, --help                 bool    prints this usage information (default: false)
  -t, --target               string  location of the Ops Manager VM
  -u, --username             string  admin username for the Ops Manager VM (not required for unauthenticated commands)
  -p, --password             string  admin password for the Ops Manager VM (not required for unauthenticated commands)
  -k, --skip-ssl-validation  bool    skip ssl certificate validation during http requests (default: false)

Command Arguments:
  -i, --installation            string  path to installation.
  -dp, --decryption-passphrase  string  passphrase for Ops Manager to decrypt the installation
```

Changes:
- Changed multipart by dragging out the adding a file functionality to a different function called AddFile(), and also introduced a new function to add form fields called AddField(). This will allow caller to add as many fields and/or files as they wish, before generating the final multipart form using Create() (which only takes the form content and generates a multipart body).
- Added Import() function to InstallationService which imports the installation using multipart form upload (the form includes the installation file and the decryption passphrase as a form field).
- Added import_installation.go to hold the execution code for the command.
- Updated main.go to create both an importInstallationService and an exportInstallationService, the former taking in an unauthenticatedClient and the latter an authedClient.


acceptance/import_installation_test.go
```diff
@@ -0,0 +1,142 @@
...
+		_, err = content.WriteString("content so validation does not fail")
+		Expect(err).NotTo(HaveOccurred())
+
+		server = httptest.NewTLSServer(http.HandlerFunc(func(w http.ResponseWriter, req *http.Request) {
+			var responseString string
+			w.Header().Set("Content-Type", "application/json")
+
+			switch req.URL.Path {
+			case "/api/v0/diagnostic_report":
+				responseString = "{}"
```
zachgersh: Why is there a diagnostic call on this test when no call is made to this endpoint?

api/installation_service.go
```diff
@@ -52,3 +60,38 @@ func (is InstallationService) Export(outputFile string) error {
 
 	return nil
 }
+
+func (is InstallationService) Import(input ImportInstallationInput) error {
```
zachgersh: We seem to be missing logic that checks as to whether the Ops Manager is already configured. If it is, this should error in a more graceful way than `422`

formcontent/formcontent.go
```diff
@@ -10,73 +10,87 @@ import (
...
+
+	return Form{
+		multipartWriter: multipart.NewWriter(body),
+		body:            body,
+	}, nil
 }
 
-func (f Form) Create(path string) (ContentSubmission, error) {
-	originalContent, err := os.Open(path)
+func (f Form) Create() (ContentSubmission, error) {
```
zachgersh: Can we rename this to something like `Finalize`? Want to make it clear that one should only call this after they are done adding all of the content in the form.

acceptance/import_installation_test.go
aditya87: Done

formcontent/formcontent.go
aditya87: Done.

api/installation_service.go
aditya87: Done.

aditya87: Can't use diagnostic service for this -- we do not provide a username and password when we import the installation, so there's no way the diagnostic service can get an authenticated client even if the ops manager is configured. I would say we ditch that and just merge this PR as is. The 422 check is good enough.

aditya87: Done...used SetupService to ensure non-availability before actually importing the installation.

PR #7, opened 2016-10-18
Show help when command is not passed

alex-slynko:
`unknown command:` was not very helpful


zachgersh:
@alex-slynko nice, I needed this! Would have been nice to add a test but I am going to merge this as is and add a test here:

https://github.com/pivotal-cf/om/blob/master/acceptance/help_test.go

PR #8, opened 2016-10-19
apply-changes now exists

zachgersh:
code review welcome


api/installations_service.go
```diff
@@ -0,0 +1,138 @@
...
+	Status string
+	Logs   string
+}
+
+func NewInstallationsService(client httpClient) InstallationsService {
+	return InstallationsService{
+		client: client,
+	}
+}
+
```
aditya87: While I understand the motivation for the installations service to only have stateless functions like Trigger, Logs, Status, and have the actual command deal with tracking the status, iterating through logs, etc...I'm just wondering if a lot of that stateful business logic should instead be part of the service? This way the command can be lightweight, and would only be responsible for passing the parameters down to the service itself. I feel like this respects objects boundaries better, but you may have a different opinion on this.

commands/log_writer_test.go
```diff
@@ -0,0 +1,53 @@
+package commands_test
+
+import (
```
aditya87: Not sure why log_writer is under commands? It almost seems like a helper (just like formcontent), and maybe it should be a separate package? I think the only things under commands should be actual commands that are available to the user to run.

main.go
```diff
@@ -14,6 +14,8 @@ import (
 
 var version = "unknown"
 
+const applySleepSeconds = 1
```
aditya87: I wonder if this can be an optional flag that we can set depending on the network latency of different IaaSes? Also is 1s a tad too frequent? Can we go to 10s?

api/installations_service.go
zachgersh: The command is in charge of orchestrating the services in what we've currently established. If I pushed the logic of retires down into each service, I would then have to have that logic coded separately for each of the three functions I have. Bringing the orchestration up to the highest level makes it more clear how the interplay of retires happens across all of the the services and that means the services do get to be mostly dumb about what is happening above them.

commands/log_writer_test.go
zachgersh: yeah, I wasn't sure I liked this either but it was so small that I wanted to leave it in commands until it really needed to be moved out. I could go either way on this and didn't love it.

main.go
zachgersh: I don't know if I want to give people the ability to tune this. Setting it higher does seem like a good idea in my mind.

zachgersh:
going to go ahead and merge this and then make the timeout change.

PR #10, opened 2016-11-01
Adding support for configurable timeouts on requests

adi-mishra:
Moved the timeouts from oauth_client.NewOAuthClient.httpclient because the Timeout variable at http.Client level won't take effect due to the way oauth2 constructs the client. And didn't want to set 'ResponseHeaderTimeout' in Transport layer because it doesn't seem to be the expected behaviour of the option provided in cli.

### DIFF
--- a/acceptance/help_test.go
+++ b/acceptance/help_test.go
@@ -19,6 +19,7 @@ Usage: om [options] <command> [<args>]
   -u, --username             string  admin username for the Ops Manager VM (not required for unauthenticated commands)
   -p, --password             string  admin password for the Ops Manager VM (not required for unauthenticated commands)
   -k, --skip-ssl-validation  bool    skip ssl certificate validation during http requests (default: false)
+  -r, --request-timeout      int     timeout in seconds for HTTP requests to Ops Manager (default: 1800)
 
 Commands:
   apply-changes             triggers an install on the Ops Manager targeted
@@ -45,6 +46,7 @@ Usage: om [options] configure-authentication [<args>]
   -u, --username             string  admin username for the Ops Manager VM (not required for unauthenticated commands)
   -p, --password             string  admin password for the Ops Manager VM (not required for unauthenticated commands)
   -k, --skip-ssl-validation  bool    skip ssl certificate validation during http requests (default: false)
+  -r, --request-timeout      int     timeout in seconds for HTTP requests to Ops Manager (default: 1800)
 
 Command Arguments:
   -u, --username                string  admin username
--- a/network/oauth_client.go
+++ b/network/oauth_client.go
@@ -19,9 +19,10 @@ type OAuthClient struct {
 	password    string
 	target      string
 	client      *http.Client
+	timeout     time.Duration
 }
 
-func NewOAuthClient(target, username, password string, insecureSkipVerify bool) (OAuthClient, error) {
+func NewOAuthClient(target, username, password string, insecureSkipVerify bool, requestTimeout time.Duration) (OAuthClient, error) {
 	conf := &oauth2.Config{
 		ClientID:     "opsman",
 		ClientSecret: "",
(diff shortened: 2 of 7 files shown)


acceptance/help_test.go
```diff
@@ -45,6 +46,7 @@ Usage: om [options] configure-authentication [<args>]
   -u, --username             string  admin username for the Ops Manager VM (not required for unauthenticated commands)
   -p, --password             string  admin password for the Ops Manager VM (not required for unauthenticated commands)
   -k, --skip-ssl-validation  bool    skip ssl certificate validation during http requests (default: false)
+  -r, --request-timeout      int     timeout in seconds for HTTP requests to Ops Manager (default: 1800)
```
aditya87: This default isn't actually enforced anywhere. Please add a test and check the default.

network/oauth_client.go
```diff
@@ -64,6 +64,7 @@ func (oc OAuthClient) Do(request *http.Request) (*http.Response, error) {
 	}
 
 	client := oc.oauthConfig.Client(oc.context, token)
+	client.Timeout =  oc.timeout
```
aditya87: This client.Timeout actually should be the same as the one that was on the `http.Client` above. Can we please add it back to that struct so that all of the relevant options at construction are grouped together?

network/oauth_client.go
```diff
@@ -39,9 +40,7 @@ func NewOAuthClient(target, username, password string, insecureSkipVerify bool)
 				Timeout:   5 * time.Second,
 				KeepAlive: 30 * time.Second,
 			}).Dial,
-			ResponseHeaderTimeout: 15 * time.Minute,
```
aditya87: We don't necessarily know how this is going to work out but we are ok with trying an overall timeout. This ResponseHeader timeout was added due to issues we were seeing on both Azure/GCP. It may be added back.

network/oauth_client.go
```diff
@@ -54,6 +53,7 @@ func NewOAuthClient(target, username, password string, insecureSkipVerify bool)
 		password:    password,
 		target:      target,
 		client:      nil,
+		timeout:     requestTimeout,
```
aditya87: No need to pass this once we move the configuration back up.

network/oauth_client.go
```diff
@@ -83,6 +84,7 @@ func (oc OAuthClient) RoundTrip(request *http.Request) (*http.Response, error) {
 	}
 
 	client := oc.oauthConfig.Client(oc.context, token)
+	client.Timeout =  oc.timeout
```
aditya87: Remove, see comment above.

zachgersh:
I cosigned this with @aditya87 - was a pair review.

network/oauth_client.go
adi-mishra: It won't be respected if it is added in http.Client, as explained in my comment for this PR. This is because in oauth2.NewClient it only inherits the Transport section of the Client thats passed in.

aditya87: No sir, it actually is added. `oauth2` implements its own transport - slightly confusing:

https://github.com/golang/oauth2/blob/master/internal/transport.go#L35-L39

network/oauth_client.go
aditya87: We added the ResponseHeaderTimeout because we wanted to make sure that we also enforced that. This is separate from the overall timeout, which, as replied below, we believe is respected. We are OK with removing it since we agree that it should cover the entire lifecycle of the connection, but if we see issues related to this we may add it back.
https://blog.cloudflare.com/the-complete-guide-to-golang-net-http-timeouts/

adi-mishra: Ok.

network/oauth_client.go
aditya87: Ah I see. OK, that makes sense now

aditya87: I'll merge this -- may do some cleanup after.

zachgersh:
@adi-mishra can you please pull the insecureSkipVerify commit off of this branch?

PR #11, opened 2016-11-01
Add Content-Type application/json to requests

bengro:
@alex-slynko observed that one can't pass JSON playload to the curl command.

The suggested fix, adds a Content-Type: application/json to all outgoing request. This fixed our issue.


aditya87:
LGTM 💯

PR #12, opened 2016-11-01
Passing through InsecureSkipVerify option in oauth client

adi-mishra:
Signed-off-by: aditya mishra <amishra@pivotal.io>


aditya87:
Nice, catch - can we add a small test for this?

zachgersh:
LGTM - test seems correct.

PR #13, opened 2016-11-03
Add live logging during upload-product

aditya87:


zachgersh:
@aditya87 I'd love to reduce the complexity of this code before we move it. The bit that is doing the live logging currently consists of what is a hard structure to comprehend. This is how it is currently laid out.

A go routine > starting a for > with a select in it > with a case statement in it > with an if inside that > with another if inside that.
 
To fix a test failure the other day, which was all timing related I had to spend about 45 minutes reasoning about the code.

Can you find a way to not have all the nesting going on and reduce this complexity?

aditya87:
Needs some work

PR #14, opened 2016-11-07
Adding retries to the oauth token retrieval call

pdelagrave:
Adding retries to the oauth token retrieval call, to avoid transient timeouts [#133285183]

Signed-off-by: Felix Reyn <freyn@pivotal.io>

### DIFF
--- a/network/oauth_client.go
+++ b/network/oauth_client.go
@@ -12,6 +12,7 @@ import (
 	"golang.org/x/oauth2"
 )
 
+
 type OAuthClient struct {
 	oauthConfig *oauth2.Config
 	context     context.Context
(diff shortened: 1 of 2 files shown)


network/oauth_client.go
```diff
@@ -96,3 +101,16 @@ func (oc OAuthClient) RoundTrip(request *http.Request) (*http.Response, error) {
...
+	count := 0
+	for (count < retryCount) {
+		token, err = oc.oauthConfig.PasswordCredentialsToken(oc.context, oc.username, oc.password)
+		if e, ok := err.(net.Error); !(ok && e.Timeout()) {
+			return
+		}
+		count += 1
+		time.Sleep(30 * time.Second)
+	}
+	return
```
pdelagrave: Does https://github.com/pivotal-cf/om/pull/10 cover these changes? Are you interested in this PR as well?

zachgersh:
@pdelagrave - going to close this PR. Talked to Felix and I believe we came to the conclusion that this doesn't actually solve your problem.

Let me know if this is wrong and I would be glad to reopen.

PR #16, opened 2016-11-15
More logging

aditya87:
Refactored live logging code to be less complex in terms of the level of branching back and forth between select-cases and ifs. Basically wrote a function to track the progress bar and write to a channel once the bar is done, and made the logging code wait on that channel to start logging (in case of tasks where the live logging happens after the progress bar is done, like import-installation and upload-product). The logging code for export-installation is unchanged.

Logically, this is much better than my previous solution, and in future the progress tracking code can theoretically be integrated into progress.go (so the code that uses the progress bar need only call Kickoff()).

### DIFF
--- a/api/products_service.go
+++ b/api/products_service.go
@@ -89,21 +89,14 @@ func (p ProductsService) Upload(input UploadProductInput) (UploadProductOutput, 
 
 	p.progress.Kickoff()
 	respChan := make(chan error)
-	progressDone := p.trackProgress()
 	go func() {
 		for {
-			var pbDone bool
-			select {
-			case _ = <-progressDone:
+			if p.progress.GetCurrent() == p.progress.GetTotal() {
 				p.progress.End()
-				pbDone = true
-			default:
-				time.Sleep(1 * time.Second)
-			}
-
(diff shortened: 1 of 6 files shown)


zachgersh:
@aditya87 got some merge conflicts here - I am still going to review. When you get a chance can you fix please?

api/products_service.go
```diff
@@ -95,14 +96,13 @@ func (p ProductsService) Upload(input UploadProductInput) (UploadProductOutput,
 			case _ = <-respChan:
 				p.liveWriter.Stop()
 				return
+			case _ = <-progressDone:
+				p.progress.End()
+				p.liveWriter.Start()
+				liveLog = log.New(p.liveWriter, "", 0)
 			default:
```
zachgersh: Does having this default case still result in a race condition? The tests end up breaking on the live-log updating because I request ends up in the `progressDone` state before ever kicking off `liveLog`. I would love for this not to be a default case and instead be checked explicitly.

aditya87: Fixed whatever race condition there was.

PR #21, opened 2016-12-07
Incorrect response to error

adi-mishra:
Wrote an integration test to show an unexpected error we saw when we were writing our integration tests.


zachgersh:
hey @adi-mishra - we wouldn't merge an acceptance test like this but it would be nice to write a unit test that covers this case.

PR #23, opened 2016-12-16
Fixed product metadata extraction filename filter

pdelagrave:
The original code was failing when an existing .yml file existed in the archive and appeared in the file list before the needed file `metadata/file.yml`. The wrong file was parsed and some required fields were missing.

For example with this product:
https://network.pivotal.io/products/push-notification-service#/releases/2285

The metadata file is expected to be found at `/metadata/*.yml` according to https://github.com/pivotal-cf/installation/blob/master/web/app/models/tempest/verifiers/internal/product_assets_verifier.rb#L70

Signed-off-by: Felix Reyn <freyn@pivotal.io>

### DIFF
--- a/extractor/product_unzipper.go
+++ b/extractor/product_unzipper.go
@@ -5,9 +5,8 @@ import (
 	"errors"
 	"fmt"
 	"io/ioutil"
-	"strings"
-
 	yaml "gopkg.in/yaml.v2"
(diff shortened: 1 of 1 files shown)


zachgersh:
@pdelagrave manually applied this. I added a test to make sure that this doesn't regress in the future.

Issue #24, opened 2016-12-20
Availability Zone configuration needs to be optional for IaaS that don't support AZ's


calebwashburn:
For Azure where OpsManager doesn't support an availability_zone edit page the current OM fails with 500 response code trying to fetch AZ guids to be used for configuration of networks subnet and network assignments.  This guid should default to `null-az`

aditya87:
Merged PR.

PR #25, opened 2016-12-20
address issue with OM working with IaaS providers that don't support …

calebwashburn:
Address issue https://github.com/pivotal-cf/om/issues/24


dave-malone:
Nice!

Issue #26, opened 2017-01-05
Bosh Director AZ configuration for vSphere requires additional fields.


malston:
vSphere requires the following fields to be filled in:

availability_zones[availability_zones][][name]
availability_zones[availability_zones][][cluster]
availability_zones[availability_zones][][resource_pool]

Where other IaaS that OM supports only uses the following field for AZ:

availability_zones[availability_zones][][iaas_identifier]

ryanmoran:
This will be resolved in these stories:
- https://www.pivotaltracker.com/story/show/136967791
- https://www.pivotaltracker.com/story/show/136633283

PR #27, opened 2017-01-05
Adding support to enable configuring AZ's for vSphere bosh director tile

malston:
This addresses the following issue: https://github.com/pivotal-cf/om/issues/26


ryanmoran:
Hey, we already have stories that were pointed this week in IPM to support this:
- https://www.pivotaltracker.com/story/show/136967791
- https://www.pivotaltracker.com/story/show/136633283

Because the solution we have landed upon is more inline with what we would expect for the `om` API (mainly, no IAAS-specific API nomenclature), we will continue ahead with these stories instead of merging this.

PR #30, opened 2017-01-10
Reuse status validation with consistent messaging

asehra:
The same HTTP status code check with similar error messaging occurred over 20 times in different service object.

### DIFF
--- a/api/installation_asset_service.go
+++ b/api/installation_asset_service.go
@@ -147,8 +147,7 @@ func (ia InstallationAssetService) Import(input ImportInstallationInput) error {
 
 	defer resp.Body.Close()
 
-	err = ValidateStatusOK(resp)
-	if err != nil {
+	if err = ValidateStatusOK(resp); err != nil {
 		return err
 	}
 
--- a/api/installation_asset_service.go
+++ b/api/installation_asset_service.go
@@ -6,7 +6,6 @@ import (
 	"io"
 	"log"
 	"net/http"
-	"net/http/httputil"
 	"os"
 	"time"
 )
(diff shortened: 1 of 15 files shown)


zachgersh:
Hey @asehra like the PR, wanted to see a very small change.

If `ValidateStatusOK` is the last call in the function can you please explicitly assign the error and return it? We like the explicitness of having the function end at only nil and not having to jump to another file to figure out where control continues.

I have marked a few of the spots that I'd like to see changed back as examples.

api/stemcell_service.go
```diff
@@ -50,14 +49,5 @@ func (us UploadStemcellService) Upload(input StemcellUploadInput) (StemcellUploa
...
-		out, err := httputil.DumpResponse(resp, true)
-		if err != nil {
-			return StemcellUploadOutput{}, fmt.Errorf("request failed: unexpected response: %s", err)
-		}
-
-		return StemcellUploadOutput{}, fmt.Errorf("request failed: unexpected response:\n%s", out)
-	}
-
-	return StemcellUploadOutput{}, nil
+	return StemcellUploadOutput{}, ValidateStatusOK(resp)
```
zachgersh: This would now turn into

```go
err = ValidateStatusOK(resp)
if err != nil {
  return StemcellUploadOutput{}, err
}

return StemcellUploadOutput{}, nil
```

api/installation_asset_service.go
```diff
@@ -148,16 +147,7 @@ func (ia InstallationAssetService) Import(input ImportInstallationInput) error {
...
-		out, err := httputil.DumpResponse(resp, true)
-		if err != nil {
-			return fmt.Errorf("request failed: unexpected response: %s", err)
-		}
-
-		return fmt.Errorf("request failed: unexpected response:\n%s", out)
-	}
-
-	return nil
+	return ValidateStatusOK(resp)
```
zachgersh: Same sort of change would happen in this function.

api/stemcell_service.go
asehra: @zachgersh Got it. 

Do you also prefer `err` assignment outside of the `if` conditional clause as in the example above? 

I have inlined the `err` assignment in other places where `ValidateStatusOk` was not in return (Eg. [here](https://github.com/pivotal-cf/om/pull/30/commits/d9f2ac2e916016353df8058afacc50635aab64b8#diff-3838a81da13a8c04458b725edd0e4045R93)). I can move them into different lines if that is the style preference.

zachgersh: @asehra inlining it is totally fine 💃

asehra: @zachgersh Great.. All done

PR #37, opened 2017-01-26
Making InternetConnected for resources optional. Fails if this is provided to a tile being deployed to vsphere

rahulkj:
Making `InternetConnected` for resources optional. Fails if this is provided to a tile being deployed to vsphere

Following is the error that's sent from the server if the InternetConnected flag is set on resources

```
{"errors":{"internet_connected":["unsupported for this infrastructure"]}}
```

After adding the `omitempty` option to the `InternetConnected`, everything worked as expected against vsphere deployment


madamkiwi:
Instead of merging directly, we've added a test in addition your change:
https://github.com/pivotal-cf/om/commit/803eba7e23216b4d1078e4c80e42d3af89b2e3bd

PR #44, opened 2017-02-21
add support for OpenStack IaaS configuration

DanielJonesEB:
We've tested this on a Pivotal Field engagement, and it seems tickety-boo against a v2 auth server and Ops Manager 1.9.3.

### DIFF
--- a/acceptance/configure_bosh_test.go
+++ b/acceptance/configure_bosh_test.go
@@ -372,6 +372,73 @@ var _ = Describe("configure-bosh command", func() {
 		})
 	})
 
+	Context("OpenStack", func() {
+		var command *exec.Cmd
+
+		BeforeEach(func() {
+			iaasConfiguration := `{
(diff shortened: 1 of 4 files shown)


zachgersh:
@DanielJonesEB - I'd like to clarify whether `tickety boo` is a superlative or a pejorative. Wanted to know so I can work this into my vocabulary.

DanielJonesEB:
@zachgersh Tickety-boo is a synonym for "hunky-dory". Hope that clears things up!

On a more serious note... Apparently it's of Hindi origin, and came back to the UK via military vocab. 

The OpenStack we're testing against has a few problems that mean we can't deploy the Director, but we've tested these changes to `om` in as much as submitting the data and Ops Manager successfully verifying the settings.

zachgersh:
LGTM - especially because we'll just call this experimental in the release notes.

PR #50, opened 2017-03-14
Add some details on how to get opsman token

z4ce:
Just updating this doc with some more detail. I've had a couple people ask me about these values in the last 24 hours, so I figure it should be in the documentation.


zachgersh:
@z4ce - wouldn't a user never need to get the ops-manager token? `om` already gets that for you as long as you provide your ops manager username / password to it. Why would you want to get the token ahead of time?

z4ce:
@zachgersh This is the readme section for setting up a Client ID an Client Secret. In that case you have to manually get the token to setup the client id/secret that OM will use the future.

PR #55, opened 2017-03-18
Provide empty errands param to delete-installation

jamiemonserrate:
OpsManager 1.10 expects an errands param, which is an array of errands to be
run before delete. Passing in an empty value as we don't want to run any
errands before destroying. This should stay consistent with current
behaviour.


utako:
@jamiemonserrate has confirmed that this fix is backwards compatible. merging now.

PR #62, opened 2017-04-11
Add --silent, -s flag to curl command

goonzoid:
If specified, http response headers won't be written to stderr unless the status code is 4XX or 5XX.

### DIFF
--- a/commands/curl.go
+++ b/commands/curl.go
@@ -27,6 +27,7 @@ type Curl struct {
 		Path   string `short:"p" long:"path"    description:"path to api endpoint"`
 		Method string `short:"x" long:"request" description:"http verb" default:"GET"`
 		Data   string `short:"d" long:"data"    description:"api request payload"`
+		Silent bool   `short:"s" long:"silent"  description:"only write response headers to stderr if response status is 4XX or 5XX"`
 	}
 }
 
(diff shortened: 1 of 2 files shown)


zachgersh:
@goonzoid - so the precent was actually important (silly I know). It was there to make sure that we if there was a percent sign in the response body that our program didn't blow up. Previously we'd been using `Printf` and go thought we were missing a variable :/

Can we add a test back for that please?

goonzoid:
Hah! Sure, I'll revert that commit. I thought it was a typo that had persisted through copy/paste or something!

goonzoid:
Done.

Nothing to do with this PR, but I wonder if it's worth calling it out that that's important somehow. I bet I won't be the last person to think it might have been a typo! Something like `response-value-with-percent-sign%` and I'd never have considered removing it.

goonzoid:
Just realised I didn't actually add a test for the negative case when the flag isn't specified. Let me know if you think that's worth doing and I'll stick it in. It's easy enough to do!

utako:
I think you're right, it'd be great to add a test for the negative case. If you can add that extra test in, then I think we're all set to merge. Thanks!

goonzoid:
Okay, I updated the top level default usage test to cover that. I can think of a few more cases that _could_ be covered, but it feels like overkill to me given the simplicity of the implementation.

I also slimmed down the new test cases a little in an attempt to reduce clutter.

Let me know if you'd like any other changes though, happy to oblige.🌟 🍰 🎩

PR #63, opened 2017-04-12
Enable om curl to support opsmanager 1.7

peterellisjones:
Ops manager 1.7 expects certain endpoints to send data encoded as `application/x-www-form-urlencode` (eg /api/v0/installations). This PR adds a new flag to the `om curl` command, `--content-type` that allows the user to override the content type header in order to send non-json encoded requests. By default it will still send requests as `application/json`

Signed-off-by: Ashish Sehra <ashish.sehra@armakuni.com>

### DIFF
--- a/api/request_service.go
+++ b/api/request_service.go
@@ -7,9 +7,10 @@ import (
 )
 
 type RequestServiceInvokeInput struct {
-	Path   string
-	Method string
-	Data   io.Reader
+	Path        string
+	Method      string
+	Data        io.Reader
+	ContentType string
 }
 
 type RequestServiceInvokeOutput struct {
@@ -32,7 +33,7 @@ func (rs RequestService) Invoke(input RequestServiceInvokeInput) (RequestService
 		return RequestServiceInvokeOutput{}, fmt.Errorf("failed constructing request: %s", err)
 	}
 
-	request.Header.Set("Content-Type", "application/json")
+	request.Header.Set("Content-Type", input.ContentType)
 	response, err := rs.client.Do(request)
 	if err != nil {
 		return RequestServiceInvokeOutput{}, fmt.Errorf("failed submitting request: %s", err)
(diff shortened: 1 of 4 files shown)


zachgersh:
hey @peterellisjones - apologies but we had one PR In front of you that was touching the same command. Would you mind resolving the conflicts on this? Glad to merge this after.

peterellisjones:
thanks @zachgersh, it's now rebased with the new commits.

zachgersh:
@peterellisjones - as we've been trying to make this command match up with actual `curl` would you be open to instead of having a `content-type` flag instead adding the `-H` flag to just attach headers?

peterellisjones:
@zachgersh we thought about that however in curl `-H` can be passed multiple times (eg: `curl -H "Accept: application/json" -H "content-type: text/html"`). The current [flag parsing methods](https://github.com/pivotal-cf/om/blob/master/flags/parse.go) would need to be re-written to support the same flag being passed multiple times.

An alternative would be to just support one `-H` header, but this might confuse users used to curl's behaviour.

zachgersh:
DIdn't mean to leave this hanging. I actually think I will end up _attempting_ to support the multiple `-H` flags at the very least here. Give me a week.

zachgersh:
@peterellisjones - this is now on master. Merged your changes and just did a quick reformat to support headers. Would love feedback if you wouldn't mind building a go binary to try this out.

peterellisjones:
Thanks Zach — that looks great. I’m no longer on the team that was using OM
but I’ve let them know the PR has been merged so they can unfork it.


On 25 April 2017 at 06:36:42, Zach Gershman (notifications@github.com)
wrote:

@peterellisjones <https://github.com/peterellisjones> - this is now on
master. Merged your changes and just did a quick reformat to support
headers. Would love feedback if you wouldn't mind building a go binary to
try this out.

—
You are receiving this because you were mentioned.
Reply to this email directly, view it on GitHub
<https://github.com/pivotal-cf/om/pull/63#issuecomment-296916064>, or mute
the thread

PR #87, opened 2017-07-06
Security and Resource config can change after bosh is deployed

asehra:
Configure-bosh command should allow updating security and resource configurations after bosh has been deployed. 

Signed-off-by: Robin Watkins <robin.watkins@hsbc.com>


acrmp:
Hi Ashish and Robin,

Thanks for taking the time to contribute this change.

It looks good, although the acceptance tests are failing because they need to be updated to reflect the different order that you are applying configuration.

Please update the acceptance tests and then we'll look at pulling this in.

Thanks,

Andrew.

asehra:
@acrmp Thanks! We've fixed it now.

acrmp:
Thank you!. I squashed the two commits and pushed as 2c53ca0f84306b47d0dc068644dc8da6aa74ee62.

PR #88, opened 2017-07-10
Add NSX Support

christianang:
Hey y'all,

This PR is to support nsx properties in `om configure-bosh` and `om configure-product`. Let us know if you have any questions/issues.

Thanks,
@christianang and @xchapter7x

### DIFF
--- a/acceptance/configure_bosh_test.go
+++ b/acceptance/configure_bosh_test.go
@@ -590,5 +590,48 @@ var _ = Describe("configure-bosh command", func() {
 			_, ok := Forms[0]["iaas_configuration[subscription_id]"]
 			Expect(ok).To(BeFalse())
 		})
+
+		Context("when adding nsx properties", func() {
+			BeforeEach(func() {
+				iaasConfiguration := `{
+					"vcenter_host": "some-vcenter-host",
+					"vcenter_username": "my-vcenter-username",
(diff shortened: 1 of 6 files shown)


acceptance/configure_bosh_test.go
```diff
@@ -570,6 +575,11 @@ var _ = Describe("configure-bosh command", func() {
 			Expect(Forms[0].Get("iaas_configuration[bosh_vm_folder]")).To(Equal("some-vm-folder"))
 			Expect(Forms[0].Get("iaas_configuration[bosh_template_folder]")).To(Equal("some-template-folder"))
 			Expect(Forms[0].Get("iaas_configuration[bosh_disk_path]")).To(Equal("some-disk-path"))
+			Expect(Forms[0].Get("iaas_configuration[nsx_networking_enabled]")).To(Equal("true"))
```
joshzarrabi: Hi @christianang. We would like to see the NXS configuration tested in a separate context. We think it will make it more clear that those are optional parameters. 

Thanks so much

Josh && @kkallday

christianang:
Updated the acceptance test for `configure-bosh` to be in a separate context. Have another look and let us know if there is anything else you want us to have another look at.

Thanks.

joshzarrabi:
Awesome, thanks so much!

PR #100, opened 2017-08-04
Adding full example for product-resources

jasper9:
It took me a lot of googling to find a mention of the full syntax for instance_type and persistent_disk for this section.   Strange the names of the parameters differ so much from the source of the web ui.  I didn't look at a metadatafile though, hopefully it matches there?

### DIFF
--- a/docs/configure-product/README.md
+++ b/docs/configure-product/README.md
@@ -152,6 +152,17 @@ For the current configuration of your product, you can `curl` the API to retriev
 
 ### Configuring the `--product-resources`
 
+```json
+{
+  "job-name": {
+    "instances": "automatic",
+    "instance_type": {"id": "automatic"},
+    "persistent_disk": {"size_mb": "automatic"}
+  }
+}
+```
+
+
 #### Example JSON:
 ```json
 {


docs/configure-product/README.md
```diff
@@ -152,6 +152,17 @@ For the current configuration of your product, you can `curl` the API to retriev
 
 ### Configuring the `--product-resources`
 
+```json
+{
+  "job-name": {
+    "instances": "automatic",
```
kkallday: I tried to set the VMs instance count with the number as a string but Ops Manager returned a `422`. When this is set to an integer is works. Although `"automatic"` works, for example purposes, it would be better to specify this as some number since `"automatic"` is already the default.

I think it would be ideal to mix-in the assignment of `instance_type` and `persistent_disk` with the example manifest below. For instance, set `mysql`'s `persistent_disk` type to `204800`, and `consul_server`'s `instance_type` to `large.mem`.

kkallday:
Hi @jasper9,

Sorry you had a lot of trouble tracking this down. The Ops Manager docs are a good resource for finding this kind of information. In particular the format is mentioned [here](https://opsman-dev-api-docs.cfapps.io/#configuring-resources-for-a-job). And you are correct, the job names match up based on what is in the metadata.

Please see the requests for changes in the PR.

docs/configure-product/README.md
kkallday: I tried to set the VMs instance count with the number as a string but Ops Manager returned a `422`. When this is set to an integer is works. Although `"automatic"` works, for example purposes, it would be better to specify this as some number since `"automatic"` is already the default.

I think it would be ideal to mix-in the assignment of `instance_type` and `persistent_disk` with the example manifest below. For instance, set `mysql`'s `persistent_disk` type to `204800`, and `consul_server`'s `instance_type` to `large.mem`.

madamkiwi:
Hi @jasper9! Just following up to see if you have any questions or concerns regarding the change feedback.

madamkiwi:
Thanks a lot for this pull request. On reflection we decided against extending the documentation to include these fields and instead [would encourage people to refer to the Ops Manager API documentation](https://github.com/pivotal-cf/om/#documentation) included with their version of Ops Manager.

Thanks,

K && Andrew

PR #106, opened 2017-09-29
Log job name before getting existing job config

christianang:
Currently `configure-product` output prints the name after it gets the existing job config. This can be confusing if you give a job name that doesn't exist and assume that the job that failed is the one that is printed.

e.g I saw this in my pipeline

```
applying resource configuration for the following jobs:
	backup-prepare
could not execute "configure-product": could not fetch existing job configuration: request failed: unexpected response:
```

om succeeded in configuring `backup-prepare`, but failed to configure `ccdb`, which was removed in pcf 1.12. Not quick to find out what was failing.


davewalter:
Looks good, thanks!

PR #110, opened 2017-10-03
switch to golang standard dep

zachgersh:
- dependency management should be done with dep
- bonus, now less dependencies are vendored due to
prune (basically removed 3x the number of files)


zachgersh:
Actually - need to look at this, the acceptance tests failed. 💃

zachgersh:
All the tests pass again biggest things are:

1. pb v2 is VERY different than v1, probably don't want to switch to that.
2. tables have more padding now, looks nice.
3. Google has significantly modified the behavior of OAuth2 (which probably should be bumped). To make the change it would seen that we would need to attach Basic Auth to our UAA requests so that the `client_id` and `client_secret` would get sent along in cases where they are needed. UAA has documentation about accepting them as basic auth. For now, pinned OAuth back to the last working version.

madamkiwi:
Thanks for the PR, we are looking into this, but will need more time to consider the implication of upgrading to golang 1.8+ in our CI

PR #111, opened 2017-10-06
Add support for Floating IPs in configure-product

oozie:
Hi team,

We are adding support for Floating IPs for Openstack.

Thanks,
Slawek and @christianang


michelleheh:
Thanks!

PR #125, opened 2017-12-20
Docs:configure-bosh:gcp Add Service Network true

Pivotal-Jayson-Hairston:
Add Service Network true


davewalter:
Thanks @Pivotal-Jayson-Hairston.

PR #138, opened 2018-03-27
Propagate http errors for the list errands call

jhvhs:
Signed-off-by: Tom Chen <tochen@pivotal.io>


michelleheh:
We manually merged it as 30295f31c. Thank you!

Issue #147, opened 2018-04-27
`configure-director` command requires a more for idempotency invocations


jtarchie:
We are working on a script that calls `configure-director` every with the same argument values.
We've after the initial invocation, that errors (from OpsMan API) occur.
These errors are related to the fact that some endpoints can be set after a director has been deployed.

For example, the `--network-assignment` argument, returns an error of:

```json
{"errors":["You cannot change the network or availability zone configuration once you have deployed"]}
```

Feature Request: When `configure-director` is called multiple times with the same argument values, it Just Works™, and moves along.

ljfranklin:
Fixed in https://github.com/pivotal-cf/om/pull/148

PR #148, opened 2018-04-27
add guard for network-assignment on configure-director

jtarchie:
This guard allows network-assignment to be called multiple times without issuing an error. This helps with idempotence when calling `configure-director`.

### DIFF
--- a/acceptance/configure_director_test.go
+++ b/acceptance/configure_director_test.go
@@ -179,6 +179,9 @@ var _ = Describe("configure-director command", func() {
 
 				w.Write([]byte(`{}`))
 
+			case "/api/v0/deployed/director/credentials":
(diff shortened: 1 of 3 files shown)


jtarchie:
Related to #147

ljfranklin:
@jtarchie nice! This has been bugging me for a while, thanks for fixing it for us 🥇

zachgersh:
Just a quick note here - this error was intentional so that users wouldn't get confused when they attempted to submit their network assignment multiple times and nothing would happen.

ljfranklin:
@zachgersh thanks for that context. The downside is CI jobs that get past `configure-director` but fail on `apply-changes` can't be re-run. Having `configure-director` exit 0 but print a log message indicating it skipped the Network step feels like the lesser of the evils.

PR #149, opened 2018-04-28
add a config flag for configure-director command

### LINKED ISSUE #139: Add `yml` file-based `--config` flag to `configure-director`
The `configure-product` command now has a `--config` flag which uses [a single, flat `yml` file](https://github.com/pivotal-cf/om/tree/master/docs/configure-product#configuring-via-file) for configuring a product tile.

It would be nice if we had something similar for the [`configure-director` command](https://github.com/pivotal-cf/om/blob/8185fdfa09c0c3f8623947a8d6caa2d4bf3ed290/commands/configure_director.go#L12-L27) (which currently [have no docs](https://www.pivotaltracker.com/story/show/156673174) 😿).

What do other folks think. Is this something that would be useful?

jtarchie:
The configure-director command takes quite a few arguments that are JSON strings.
This can make it difficult to use from the command line.
The --config/-c option allows an equivalant YAML file to be used in place of the other flags.
It will return the same errors message as if the flags were invoked directly.

[Closes #139]

### DIFF
--- a/acceptance/configure_director_test.go
+++ b/acceptance/configure_director_test.go
@@ -34,6 +34,11 @@ var _ = Describe("configure-director command", func() {
 	)
 
 	BeforeEach(func() {
+		azPutCallCount = 0
+		directorPropertiesCallCount = 0
+		networkAZCallCount = 0
+		networksConfigurationCallCount = 0
+
 		server = httptest.NewTLSServer(http.HandlerFunc(func(w http.ResponseWriter, req *http.Request) {
 			w.Header().Set("Content-Type", "application/json")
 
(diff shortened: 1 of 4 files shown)


acceptance/configure_director_test.go
```diff
@@ -302,6 +308,131 @@ var _ = Describe("configure-director command", func() {
 			"internet_connected": true,
 			"elb_names": ["my-elb"]
 	  }`))
+		})
 	})
 
+	Context("when specifying a config file", func() {
+		It("configures the BOSH director using the API", func() {
+			configJSON := []byte(`
```
mcwumbly: Minor quibble. At the acceptance test level, I'd prefer to see something closer to canonical usage as it also serves as executable documentation for the app. For example, the same thing in the `configure_product` case, [here](https://github.com/pivotal-cf/om/blob/master/acceptance/configure_product_test.go#L68-L96).

Just my 2 cents!

mcwumbly:
Thanks for this! Looks good to me at a quick glance!  

Will have someone else take a look early next week :)

acceptance/configure_director_test.go
jtarchie: Honestly, I was following the example used above with the standard flags. Happy to work with someone on that canonical example. I wouldn't want it to stop this PR, though.

mcwumbly: > I wouldn't want it to stop this PR, though.

Yep, agreed. 

I think we're still in the early days of figuring out how we handle PRs effectively. Hoping for things like this, if the team has opinions they can just fix them up on the merge and then provide the feedback in a way that helps guide future contributions. 

And following existing patterns seems like a good recommendation for contributors, so we should try to keep those updated with our opinions as they change to help guide others.

commands/configure_director_test.go
```diff
@@ -117,6 +118,132 @@ var _ = Describe("ConfigureDirector", func() {
...
+					configFile, err := ioutil.TempFile("", "config.yaml")
+					Expect(err).ToNot(HaveOccurred())
+					_, err = configFile.WriteString(configYAML)
+					Expect(err).ToNot(HaveOccurred())
+					Expect(configFile.Close()).ToNot(HaveOccurred())
+
+					err = command.Execute([]string{
+						"--config", configFile.Name(),
+					})
+					Expect(err).To(HaveOccurred())
```
ljfranklin: @jtarchie minor nitpick, could we assert that a more specific error was received here? E.g. `MatchError` or `Expect(err.Error()).To(ContainSubstring(...))`. Just making sure we're hitting the error branch we expected to.

acceptance/configure_director_test.go
ljfranklin: @jtarchie not sure if this was what Dave was hinting at, but could we make this provided config file fixture string YAML rather than JSON? Little easier to read in the test as we don't have to escape stuff and it more closely matches how we expect the command to be used.

commands/configure_director_test.go
```diff
@@ -117,6 +118,132 @@ var _ = Describe("ConfigureDirector", func() {
 			Expect(logger.PrintfArgsForCall(11)).To(Equal("finished configuring resource options for bosh tile"))
 		})
 
+		Context("when the --config flag is set", func() {
+			Context("when other flags are set", func() {
+				It("returns an error", func() {
+					err := command.Execute([]string{"--config", "test.yml", "--az-configuration", "{}"})
```
ljfranklin: I like this "table-driven" style test. We should do more of this sort of thing going forward.

Issue #153, opened 2018-05-07
specify target and authorization flags via a config file


fredwangwang:
right now if we want to target the om to use specific commands, we need to provide `--target`, `--username`, `--password`, and `--skip-ssl-validation` flags each time. This became a problem when we were trying to externalize the configurations in a Concourse task. It would be nice to have a config file for target and authentication credentials so that we can pass a single config file each time instead of the same params.

And we are willing to make a PR for this issue.

ryanmoran:
What do you feel the advantage of a file would be over supporting environment variables for flags? Do you explicitly want your credentials written to disk?

fredwangwang:
Yes, we want to have a versioned control of the configurations for different foundations. It would quickly become unmanageable if we are passing in the variables by flag or env variables. So having a config file stored somewhere (like s3 bucket) provides us a central and versioned configuration we could rely on.

zachgersh:
There are quite a few ways that I could envision doing this that do not involve writing the credentials down to disk. `om` is trying to make sure you do what is the "secure" thing instead of leaving your credentials around in various files.

One such solution would be to have a file with a bunch of export statements in it that set your credentials once sourced.

mcwumbly:
closing per: https://github.com/pivotal-cf/om/pull/154#issuecomment-387874050

Issue #140, opened 2018-04-11
Support template values in the config file for configure-product


mcwumbly:
The `configure-product` command now has a `--config` flag which uses [a single, flat `yml` file](https://github.com/pivotal-cf/om/tree/master/docs/configure-product#configuring-via-file) for configuring a product tile.

It would be nice if this supported template variables along the lines of what Concourse and BOSH do.  This would allow 

For example, Concourse supports templated variables that can be supplied like this:

```bash
$ fly set-pipeline --help
...
[set-pipeline command options]
...
      -v, --var=[NAME=STRING]       Specify a string value to set for a variable in the pipeline
      -y, --yaml-var=[NAME=YAML]    Specify a YAML value to set for a variable in the pipeline
      -l, --load-vars-from=         Variable flag that can be used for filling in template values in configuration from a YAML file
```

Where BOSH has similar features:

```bash
$ bosh interpolate --help
...
[interpolate command options]
      -v, --var=VAR=VALUE        Set variable
          --var-file=VAR=PATH    Set variable to file contents
      -l, --vars-file=PATH       Load variables from a YAML file
          --vars-env=PREFIX      Load variables from environment variables (e.g.: 'MY' to load MY_var=value)
          --vars-store=PATH      Load/save variables from/to a YAML file
```

Having the ability to specify template variables in om's config file would allow a more standardized way to pass in things that currently come from terraform outputs in many people's workflows, for example.

What do other folks think? Is this something that would be useful?  If we do something like this, what syntax would be preferred for the variables within the template YML files?  Which method of passing vars would be most useful to have first (command line flags, environment variables, a separate file, etc.)?

jtarchie:
What is the reason to support variables? For concourse and bosh the abstraction of variables is so that a YAML file can be reused for different deployments, respectively.

mcwumbly:
@jtarchie there are two main motivations for this in my mind:

1. separating secrets from the config template
2. decoupling the config template from the source of the values that often populate it (like terraform outputs).

I think (2) is actually the primary thing I had in mind when writing this.

It would allow you to have do something like that following (pseudo):

```sh
terraform apply
terraform output -json | yj -r > tf-outputs.yml
om configure-product -c config.yml -l tf-outputs.yml
```

If the name of a terroform output changed from `foo` to `bar`, then I would just update by `config.yml` template to use that name instead:

```diff
 some-property:
-  some-key: ((foo))
+  some-key: ((bar))
```

[...]

PR #154, opened 2018-05-07
added the target-config-file flag for om global settings

fredwangwang:
This is the PR related to issue https://github.com/pivotal-cf/om/issues/153

Signed-off-by: David Timm <dtimm@pivotal.io>

### DIFF
--- a/acceptance/help_test.go
+++ b/acceptance/help_test.go
@@ -21,6 +21,7 @@ Usage: om [options] <command> [<args>]
   --request-timeout, -r      int     timeout in seconds for HTTP requests to Ops Manager (default: 1800)
   --skip-ssl-validation, -k  bool    skip ssl certificate validation during http requests (default: false)
   --target, -t               string  location of the Ops Manager VM
+  --target-config-file, -tc  string  YAML file specifying om global parameters
   --trace, -tr               bool    prints HTTP requests and response payloads
   --username, -u             string  admin username for the Ops Manager VM (not required for unauthenticated commands, $OM_USERNAME)
   --version, -v              bool    prints the om release version (default: false)
@@ -82,6 +83,7 @@ Usage: om [options] configure-authentication [<args>]
   --request-timeout, -r      int     timeout in seconds for HTTP requests to Ops Manager (default: 1800)
   --skip-ssl-validation, -k  bool    skip ssl certificate validation during http requests (default: false)
   --target, -t               string  location of the Ops Manager VM
+  --target-config-file, -tc  string  YAML file specifying om global parameters
   --trace, -tr               bool    prints HTTP requests and response payloads
   --username, -u             string  admin username for the Ops Manager VM (not required for unauthenticated commands, $OM_USERNAME)
   --version, -v              bool    prints the om release version (default: false)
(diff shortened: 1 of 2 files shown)


ljfranklin:
@fredwangwang @jtarchie main slack convo is here: https://pivotal.slack.com/archives/C5V956L13/p1525794596000281. Copying my response from there:

RelEng here, looking at your PR: https://github.com/pivotal-cf/om/issues/153. I agree with Ryan here, I don't really want to merge that PR as I worry it encourages people to check secrets into version control rather than using more ephemeral creds from credhub/vault/whatever. As @zachgersh noted in the issue, you can work around this by checking in a script with a bunch of `export OM_PASSWORD...` lines in it, but our recommendation would be to not write creds to disk at all.

ljfranklin:
@mcwumbly I'm leaning towards closing this without merging but I'd appreciate a sanity check from you. There's probably something we could be doing to better address credential management with `om` but encouraging writing creds to disk feels like a step in the wrong direction.

aegershman:
Unsolicited, but for anyone reading this in need of ideas, what I like to do is keep my credentials files in a secure location outside the git repo entirely && I load them as env vars for the script to consume. For example, if I have some script called `myscript.sh`, I'd load in creds like this:

```bash
#!/bin/bash

# Declare needed variables. Or don't. Whatever. Your choice.
USERNAME=""
SECRET=""
TARGET=""
API_TOKEN=""
source ~/projects/credentials/myscript-creds-loader.sh

# code goes here blah blah blah
```

where `myscript-creds-loader.sh` looks like this:
```bash
#!/bin/bash
export USERNAME="actual-username"
export SECRET="actual-secret"
export TARGET="actual-value"
export API_TOKEN="actual-token"
```

Or instead of hardcoding them, you can have the `creds-loader.sh` load it from some external secure store:
```bash
export USERNAME=$(credhub get -n /concourse/mysecret --output-json | jq -r '.value.username')
export SECRET=$(credhub get -n /concourse/mysecret --output-json | jq -r '.value.password')
# etc etc
```

jtarchie:
@ljfranklin and @zachgersh, are use case is that people would be getting these values and generating a temporary YAML file.

The argument of having a config file having credentials actively promotes checking may enable it, but doesn't actively enforce it. If we can provide mechanisms and good practices to our users, so they don't do that, that would be just as effective.

The counter-argument to that is the config file that can be provided already for `configure-director` and `config-product` can have credentials in them -- be they certs, SSH keys, etc. For example, the director config for `iaas-configuration` on GCP has the service account JSON it can set.

We can try to protect the user from themselves, but we know through experience, to quote famous mathematician and chaostician Ian Malcom, "Life will find a way."

If you would still like to close this. Maybe we can better articulate our use case, from a customer experience, to discuss further.

mcwumbly:
We are going to pull this other PR into jhanda soon, and then bump it in `om`:

https://github.com/pivotal-cf/jhanda/pull/2

That should solve the immediate concern, as you can statefully set your environment.  For the other use cases being discussed here, we want to support templating variables (and secrets), a la: https://github.com/pivotal-cf/om/issues/140

So, closing this and the related issue for now.

PR #171, opened 2018-05-30
adding ability to decorate the networking configuration with guid

calebwashburn:
Similar to az configuration added support to allow updating the configuration with the same guid for each network based on network name.


heycait:
Hey @calebwashburn thanks for the PR this was story we were also looking at 😄. We were wondering if you considered the nested subnets and their guids. 

From our understanding, the guid matching should also be done for the subnets but we weren't sure what test cases you covered and whether or not this was actually a problem?

calebwashburn:
@heycait In my testing there were no side-effects if the guid wasn't carried forward.   As nothing binds specifically to the subnet names in downstream tiles.  I would be happy however to update the PR to ensure subnet guids are also mapped if so desired.  Let me know

heycait:
Thanks @calebwashburn!

PR #174, opened 2018-06-05
Staged config now handles collections

fredwangwang:
If there was a credential in a collection, we did not subsitute a placeholder in for that credential.
Did some minor refactors to make the go more idiomatic

https://www.pivotaltracker.com/story/show/157747884


fredwangwang:
For example:
Before the process
```yaml
.properties.networking_poe_ssl_certs:
    value:
    - certificate:
        configurable: true
        credential: true
        optional: false
        type: rsa_cert_credentials
        value:
          cert_pem: |
            -----BEGIN CERTIFICATE-----
            -----END CERTIFICATE-----
          private_key_pem: '***'
```
We still see the creds as "***" because the original logic doesn't handle collection types at all.

heycait:
Hi @fredwangwang thanks for the PR. Did you guys test using the output of this command to `configure-product`? We weren't sure if this should succeed wanted to double-check with you guys first.

jtarchie:
We did test it.

Are you seeing an issue?

acrmp:
@fredwang @jtarchie Thanks for getting back to us.

We were getting errors with `configure-product` for collection items with these changes. It seems the `value` is only necessary at the top level of the collection and not for each item in the collection.

I pushed something that worked for me to a wip branch here: https://github.com/pivotal-cf/om/commit/6bf20f7bb14d160c3d696a32ce99f13355c71d62

Can you take a look and we can pair tomorrow if that's helpful?

Thanks,

Andrew.

fredwangwang:
Yes you are right, the inner `value` key has to be removed.

We did the test based on the original returned data schema, which looks like:
```json
{
    "type": "collection",
    "configurable": true,
    "credential": false,
    "value": [
        {
            "name": {
                "type": "string",
                "configurable": true,
                "credential": false,
                "value": "dummy encryption key 1",
                "optional": false
            }
        }
    ]
}
```
which actually contains the inner `value` key. Some we assumed the payload schema used by `configure-product` is exactly the same (and indeed they are, except for this one somehow...)

So thanks for catching that! I think you can just merge your [WIP](https://github.com/pivotal-cf/om/commit/6bf20f7bb14d160c3d696a32ce99f13355c71d62) in @acrmp

madamkiwi:
Thanks!

PR #181, opened 2018-06-11
change placeholder separate to _

### LINKED ISSUE #178: placeholder name is not interpolatable
For `collection` types in `staged-config` command, we used the format 

`.properties.some-collection[Idx].inner-prop`, sadly it doesn't work for bosh interpolation.  The regex in bosh, which is \(\((!?[-/\.\w\pL]+)\)\), doesn't support [ & ] in variable name.

So it would be a simple fix to change from:
https://github.com/pivotal-cf/om/blob/638511cffd6b6adccffff461d4c9a60530b18d20/commands/staged_config.go#L186

To: 
```go
 innerValueNamePrefix := name + strconv.Itoa(index) + "." + innerKey.(string)
```

An example:
output snippet:
```yaml
.properties.networking_poe_ssl_certs:
  - certificate:
      cert_pem: ((.properties.networking_poe_ssl_certs0.certificate.cert_pem))
      private_key_pem: ((.properties.networking_poe_ssl_certs0.certificate.private_key_pem))
```

vars-file:
```yaml
.properties.networking_poe_ssl_certs0.certificate:
  cert_pem: hello
  private_key_pem: world
```

Generated:
```yaml
.properties.networking_poe_ssl_certs:
- certificate:
    cert_pem: hello
    private_key_pem: world
```

jtarchie:
This follows the pattern of cf-deployment when generating parameter names. These names enforce that the vars can come from separate sources without conflict.

[#157747884]

This changes the exported properties to now be named `properties_some_name.private_key` instead of `.properties.some_name.private_key`.

This is to resolve the interpolation [issue](https://github.com/cloudfoundry/bosh-cli/issues/437) with vars-file not merge objects in YAML. 

*edit:*
We also noticed that collections of credentials did not accurately retrieve the credential value when using `--include-credentials`. We added another commit to resolve that issue.

This helps resolve #178.

### DIFF
--- a/api/staged_director_service.go
+++ b/api/staged_director_service.go
@@ -31,8 +31,8 @@ type AvailabilityZonesOutput struct {
 }
 
 type AvailabilityZoneOutput struct {
-	Name     string                 `yaml:"name"`
-	Clusters []ClusterOutput        `yaml:"clusters,omitempty"`
+	Name     string          `yaml:"name"`
+	Clusters []ClusterOutput `yaml:"clusters,omitempty"`
 }
 
 type ClusterOutput struct {
--- a/commands/staged_config.go
+++ b/commands/staged_config.go
@@ -92,7 +92,9 @@ func (ec StagedConfig) Execute(args []string) error {
 		if property.Type == "selector" {
 			selectorProperties[name] = property.Value.(string)
 		}
-		output, err := ec.parseProperties(productGUID, name, property)
+		output, err := ec.parseProperties(productGUID, propertyName{
+			prefix: name,
+		}, property)
 		if err != nil {
 			return err
 		}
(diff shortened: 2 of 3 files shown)


jtarchie:
We fixed the failing test on this build. It should go through now.

fredwangwang:
There is a minor conflict (due to the PR to address iaas_configuration_guid we had earlier).

It should be simple as replacing `api/staged_director_service.go:33` from:
```go
type AvailabilityZoneOutput struct {
	Name                  string          `yaml:"name"`
	Clusters              []ClusterOutput `yaml:"clusters,omitempty"`
}
```

to:
```go
type AvailabilityZoneOutput struct {
	Name                  string          `yaml:"name"`
	Clusters              []ClusterOutput `yaml:"clusters,omitempty"`
	IAASConfigurationGUID string          `yaml:"iaas_configuration_guid,omitempty"`
}
```

madamkiwi:
Merged manually, thanks!

mcwumbly:
@jtarchie was just catching up on acceptance for https://github.com/pivotal-cf/om/pull/172 and noticed that this chance to replace `.` with `_` only got made for `staged-config` and not for `config-template`.  should we create an issue for that to follow on this one?

PR #184, opened 2018-06-13
fix order of vars-file interpolation

jtarchie:
The order of vars files are evaluated in a right to left precedence in
`bosh int`.

For example,

`manifest.yml`
```yml
a: ((name))
```

`vars1.yml`
```yaml
name: Bob
```

`vars2.yml`
```yaml
name: Moe
```

From a terminal windows far away.
```
bosh int manifest.yml -l vars1.yml -l vars2.yml
a: Moe
```

A value in `vars2.yml` will override a value defined in `vars1.yml`.

This behaviour has been added to `om int` to maintain usage patterns.

[#158312710]


michelleheh:
Merged. Thank you!

PR #187, opened 2018-06-14
Add shasum

jtarchie:
This allows a --shasum arugument to perform a shasum comparison of either a stemcell or a product trying to be uploaded.
If they don't match, then the command will fail.

Example usage:

```
om upload-product -p product.pivotal --shasum 12390812489072345iuehfge9845uy73489572345
om upload-stemcell -p stemcell.tgz --shasum 12390812489072345iuehfge9845uy73489572345
```


michelleheh:
Thanks, all!

PR #195, opened 2018-06-24
Refactor `apply-changes` and `delete-installation`

fredwangwang:
Original implementation attempts to fetch the log data from `/api/v0/installations/:id/logs` per 10 seconds (hard-coded as applySleepSeconds). It works but is not efficient.

Some drawbacks I can think of:
* Unnecessary consumption of network  (by issuing the request periodically)
* Increase the load of the opsman server (by responding to request periodically)
* Lag in log content fetching (by fetching per 10 second)
* Slow when testing (the logic will sleep 10 second each time it fetches from fake server, and it fetches 3 times for some test, which adds 30 seconds to a single test case.)

-----------------
In this PR, I changed the logic to fetch log from `/api/v0/installations/current_log`, which is an API endpoint to stream server sent events (SSE). Because of the nature of SSE, `om` only fetch one copy of the log in total, which cuts the data usage by a huge amount. 
Here are some advantage of this implementation:
* Huge reduction in network consumption 
   ~45MB vs ~500 KB, assume: linear log data growth, 30 min install duration, 500KB final log size.
  **Note**: I didn't record the data, but caculated the result given the assumption (The assumptions are made based on real PAS installaion log)
* Reduction in server api load
* Instantaneous log fetch
* Much faster testing (reduced from 1m36s to 24s)

There is no obvious functional change from the end user perspective, but it really helps in terms of data usage and more importantly, **testing speed**.


zachgersh:
When did this current_log API endpoint come into existence? If it is fairly
new did you consider backwards compatibility?

mcwumbly:
@fredwangwang up to you. if you want to add the logic to have this work for 2.1+ only, I think that'd be fine.

fredwangwang:
@mcwumbly Thanks for the reply! I would really want this code integrated into `om`, and I've spent some time to make it compatible with pre 2.1 by falling back to use `/api/v0/installations/:id/*` endpoints. Eventually, I got it working. But after it is done, I kinda don't like the code anymore...

Because:
1. If I want to keep the overall API interface simple, I ended up adding a bunch of logic into a single API call, which breaks the philosophy to keep API layer as simple as a translation to the content of the payload.
1. If I still keep the API layer simple, then the logic just spill to the implementation of `apply-changes`, which would end up with some convoluted if statements and handler for different log streams.

As I don't see a clean way to integrate it, I guess I am going to drop this PR.

However, as I said in the orignal comment, one of the main motivation I have this PR is it really speeds up the testing. A follw up question would be: Is variablize `applySleepSeconds` acceptable?

Changes:
main.go
```go
var applySleepSecondsStr = "10"

func main() {
	...
	applySleepSeconds, err := strconv.Atoi(applySleepSecondsStr)
		if err != nil {
	stdout.Fatal(err)
}
```

and

acceptance/init_test.go
```go
...
pathToMain, err = gexec.Build("github.com/pivotal-cf/om", "-ldflags", "-X main.applySleepSecondsStr=0")
...
```
Which still speeds up the testing (down to 37 seconds), but doesn't change the overall program behavior.

mcwumbly:
Thanks for the update @fredwangwang.

I'd encourage us not to bail on this effort, even if we don't solve it right away. I think it may be  worthwhile to see this as an _opportunity_ to establish a pattern for handling this type of problem as I imagine we'll encounter it again when newer features become available in Ops Manager or older features become deprecated or are removed.

I been writing much code for the past 6 months, so take my suggestion with a grain of salt, but what I imagine doing to address your concerns is something like the following:

1. have a small API object that just does version detection (or maybe feature detection).
2. use that early in any given command to know what kind of backend you're dealing with
3. based on the output of that, use a factory to create the appropriate API object to do the work
4. use that created object to do the work.

zachgersh:
Agree with @mcwumbly - you should throw this away since this is something that will eventually need to be tackled at some point in `om`. You should totally try to wrap up both sets of installation fetching logic into a higher level package like we've done with extractor and formcontent.

PR #203, opened 2018-07-13
add instnace service account key option

datianshi:
as title

### DIFF
--- a/commands/bosh_form_models.go
+++ b/commands/bosh_form_models.go
@@ -7,9 +7,10 @@ type CommonConfiguration struct {
 }
 
 type GCPIaaSConfiguration struct {
-	Project              string `url:"iaas_configuration[project],omitempty" json:"project"`
-	DefaultDeploymentTag string `url:"iaas_configuration[default_deployment_tag],omitempty" json:"default_deployment_tag"`
-	AuthJSON             string `url:"iaas_configuration[auth_json],omitempty" json:"auth_json"`
+	Project                 string `url:"iaas_configuration[project],omitempty" json:"project"`
+	DefaultDeploymentTag    string `url:"iaas_configuration[default_deployment_tag],omitempty" json:"default_deployment_tag"`
+	AssociateServiceAccount string `url:"iaas_configuration[associated_service_account],omitempty" json:"associated_service_account"`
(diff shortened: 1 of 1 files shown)


jtarchie:
@datianshi, I believe these are purposely being deprecated in favour of using `configure-director` instead of `configure-bosh`. Someone from the `om` team could confirm.

Are you able to switch commands?

datianshi:
@jtarchie  do not think this is related to two different commands at all. This is for IAAS specific configuration. Now 2.1 Ops Manager is able to configure service account instance vs using service key.

mcwumbly:
@datianshi just bumping this to see if you can help with any of my questions above. thanks!

datianshi:
It is in the doc. Just do a search key word 'associated_service_account' @mcwumbly .... That is in the example field though.... Probably the right team should further document that

mcwumbly:
Thanks for reporting this!  

Fixed via https://github.com/pivotal-cf/om/commit/92ea225d022918ef4549281359739ec72f64f723 to add tests/docs and fix a minor typo.

datianshi:
@mcwumbly  This is not a report. This is a pull request. I do not think it is a good open source practice to discard committers' commits and commits by core engineers. It is always good to embrace more contributors and advice what needs to enhance in the existing commits

mcwumbly:
Sorry for the mis-communication, @datianshi.  I did not mean at all to diminish the contribution, and didn't want to hold up your changes just because of a single-character typo and the fact that [these changes](https://github.com/pivotal-cf/om/pull/203#issuecomment-405438108) had not been explicitly addressed.

We do note the following in the [CONTRIBUTING.md](https://github.com/pivotal-cf/om/blob/master/CONTRIBUTING.md) docs now:

> When a maintainer accepts your changes, they will merge your pull request. If there are outstanding requests for changes or other small changes they feel can be made to improve the changed code, they may make additional changes or merge the changes manually. It's always nice to have changes come in just as the team would like to see them, but we'll try not to hold up a pull request for a long period of time due to minor changes.

I thought I was doing that in spirit and that referencing your PR was due attribution, but I can see why it would have been better to include your commit directly and make the fixes as a separate commit on top of that.

Do you think that would that have been better practice here?

PR #206, opened 2018-07-18
Add --product-name flag to apply-changes command

### LINKED ISSUE #141: Add "selective deploy" functionality for PCF >2.1
Add "selective deploy" functionality for PCF >2.1.

This is new coming down the pipe with 2.1 and would make om-linux and the pipelines that call in a lot cleaner and faster.

image

zmb3:
Add support for OM 2.2's selective apply changes.

Fixes #141

For example:
```
$ om  [...] apply-changes --product-name ert --product-name p-healthwatch
```

When `--product-name` is used, a call to the `info` endpoint is made to ensure that the specified ops manager supports this operation.  If it doesn't, the user receives an error:

```
$ om -t https://opsman.example.com -p redacted -u admin apply-changes --product-name p-healthwatch
could not execute "apply-changes": --product-name is only available with Ops Manager 2.2 or later: you are running 2.1-build.326
```

### DIFF
--- a/api/info_service.go
+++ b/api/info_service.go
@@ -1,0 +1,34 @@
+package api
+
+import (
+	"encoding/json"
+	"fmt"
+	"net/http"
+)
+
+// Info contains information about Ops Manager itself.
+type Info struct {
+	Version string `json:"version"`
+}
+
+// Info gets information about Ops Manager.
+func (a Api) Info() (Info, error) {
+	var r struct {
+		Info Info `json:"info"`
+	}
+	req, err := http.NewRequest("GET", "/api/v0/info", nil)
+	if err != nil {
+		return r.Info, err
+	}
+	resp, err := a.client.Do(req)
+	if err != nil {
+		return r.Info, fmt.Errorf("could not make request to info endpoint: %v", err)
+	}
+	defer resp.Body.Close()
+
+	if err = validateStatusOK(resp); err != nil {
+		return r.Info, err
+	}
+	err = json.NewDecoder(resp.Body).Decode(&r)
+	return r.Info, err
+}
--- a/api/info_service.go
+++ b/api/info_service.go
@@ -4,11 +4,35 @@ import (
 	"encoding/json"
 	"fmt"
 	"net/http"
+	"strconv"
+	"strings"
 )
 
 // Info contains information about Ops Manager itself.
 type Info struct {
 	Version string `json:"version"`
+}
+
+func (i Info) VersionAtLeast(major, minor int) bool {
+	// Given: X.Y-build.Z
+	// Extract X and Y
+	idx := strings.Index(i.Version, ".")
+	majv := i.Version[:idx]                                  // take substring up to '.'
+	minv := i.Version[idx+1 : strings.Index(i.Version, "-")] // take substring between '.' and '-'
+
+	maj, err := strconv.Atoi(majv)
+	if err != nil {
+		panic("invalid version: " + i.Version)
+	}
+	min, err := strconv.Atoi(minv)
+	if err != nil {
+		panic("invalid version: " + i.Version)
+	}
+
+	if maj < major || (maj == major && min < minor) {
+		return false
+	}
+	return true
 }
 
 // Info gets information about Ops Manager.
--- a/api/info_service_test.go
+++ b/api/info_service_test.go
@@ -12,46 +12,67 @@ import (
 	"github.com/pivotal-cf/om/api/fakes"
 )
 
-var _ = Describe("Info", func() {
-	var (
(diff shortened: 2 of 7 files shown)


jtarchie:
The `om` CLI maintain backwards compatible with older versions of OM. This API doesn't work for <2.2, right?

@mcwumbly is there some way to support new APIs while maintaining backwards compatibility?
* If someone uses a newer feature, the CLI detects the OM version, and errors that the feature is not supported?
* Having version specific commands. For example, this would be `om apply-changes-2.2`.
* ???

davewalter:
Checking the OM version adds a certain amount of overhead, but it's definitely the cleanest solution. I suspect that we are going to run into this more and more over time.

zmb3:
So what do y’all recommend we do if it’s an older version of ops manager?  Ignore the flags and do a standard apply changes, or fail with an error?

davewalter:
Personally, I think we would fail with an error if the operation can't be performed as requested.

mcwumbly:
> Personally, I think we would fail with an error if the operation can't be performed as requested.

agreed.

> Having version specific commands. For example, this would be om apply-changes-2.2

let's not do that.

> If someone uses a newer feature, the CLI detects the OM version, and errors that the feature is not supported?

this is fine, and I think we do it elsewhere, so probably a safe default

> ???

The other option is some other form of _[feature detection](https://en.wikipedia.org/wiki/Feature_detection_(web_development))_ that does not rely on the version number. This is arguably more robust, as it'd automatically start working if the feature is added in a later patch version, but I'm not sure we have any precedent yet for doing this in our code yet.

zmb3:
OK, so to recap:

> This API doesn't work for <2.2, right?

Right, where "doesn't work" simply means it behaves like a standard apply changes (there are no errors or other failures).

> Having version specific commands. For example, this would be om apply-changes-2.2

Totally agree, let's _not_ do this.  Sounds like it will set a really bad precedent.

Planning to do as @mcwumbly suggests and make a call to get the version number, raising an error if `--product-name` is used on an Ops Manager < 2.2.

zmb3:
Is it safe to assume that the version is always of the form `vX.Y-build.Z`?

mcwumbly:
> Is it safe to assume that the version is always of the form vX.Y-build.Z?

I'm not sure... I think their schema may have changes at some point from `X.Y.Z` or `X.Y.Z-build.N` to `X.Y-build.Z`

I'm kind of surprised there's no existing case already where we compare the Ops Manager version, but I can't find one...

Perhaps it's worth extracting that method and testing it separately as I imagine it will be useful in other places, given the discussion above.

davewalter:
> I'm kind of surprised there's no existing case already where we compare the Ops Manager version, but I can't find one...

The other case that I can think of where this happens is `configure-director`, which generates an error as the API endpoint simply doesn't exist before a certain version (1.12?).

> Perhaps it's worth extracting that method and testing it separately as I imagine it will be useful in other places, given the discussion above.

Agreed. It would be nice to error gracefully instead of displaying a 404 for a non-existent API endpoint.

mcwumbly:
> Happy to do this, will push an update this evening.

Sounds good!

> don't you think that should be a separate issue that's not part of this PR?

Yeah, that's fine. Also, at this point, I think we will happily merge this PR as-is when it gets to the top of our list. In the mean time, any other improvements are welcome, but not required.

doddatpivotal:
selective deploy is available in 2.1 via the UI /debug/selective_deploy.  Is it possible to allow this command to work with 2.1?

mcwumbly:
@doddatpivotal good question. Perhaps this would "just work" with 2.1 in that case?

doddatpivotal:
Well, the question is whether the URL is different for 2.1 vs 2.2.  I'm not sure the 2.2 url, does it include the path /debug/.  Also, it's not available for < 2.1.  I'm particularly interested in this for concourse pipelines focused on specific tiles.

zmb3:
Currently hard coded to fail if it detects less than 2.2. If this API is identical and does work in 2.1 we would need to make a quick change. Who can confirm?

mcwumbly:
@sadvani can you or someone else on the ops manager team weigh in here?

mcwumbly:
> If this API is identical and does work in 2.1 we would need to make a quick change.

We can certainly make this change in a follow up PR or commit once we confirm.

Also, just want to re-iterate the above:

> at this point, I think we will happily merge this PR as-is when it gets to the top of our list. In the mean time, any other improvements are welcome, but not required.

zmb3:
Affirmed. I have no further plans for this. Please merge when you get a chance - I have customers eager to use this.

Issue #214, opened 2018-07-24
Include the ability to set errand state in configure product command


calebwashburn:
As a consumer I would like to be able to call configure product to also set the errand state instead of having to call additional command `set-errand-state`.  Like resource configuration proposing that another yaml element be added to configure product yaml input to allow setting the state of each errand.  Format would be like the following:

```
errand-config:
  smoke_tests:
    post-deploy-state: enabled
    pre-delete-state: enabled
  push-usage-service:
    post-deploy-state: disabled
    pre-delete-state: disabled
```

This would allow configuration of all aspects of a given product in a single configuration command which is orchestrated by `om` to correct endpoints on Operations Manager.

chendrix:
If this section is empty when someone runs `configure-product`, I assume it doesn't change the state of the errands?

If someone runs `set-errand-state` and then runs `configure-product`, I assume the configuration in `config.yml` overrides whatever they set when they ran `set-errand-state`?

calebwashburn:
Correct @chendrix this is last in wins.  And if there is no errand-config in the config file no errand state will be updated.  Also, only errands listed will be updated.

calebwashburn:
This has been merged so closing issue.

PR #217, opened 2018-07-25
added ability to configure errand state in configure product.

calebwashburn:
Added ability to configure state of errand config in `configure-product` command.

Addresses issue - https://github.com/pivotal-cf/om/issues/214


genevieve:
@calebwashburn Do you think it makes sense to continue to support `om set-errand-state` if you can configure errand state for multiple errands for a single product now with `om configure-product`?

calebwashburn:
I think this could deprecate the `om set-errand-state` command.

Issue #204, opened 2018-07-17
Add flag to consume environment files


nbconklin:
In order to achieve the outcome of consuming om directly in the concourse tasks (as outlined [here](https://github.com/pivotal-cf/om/issues/163)), we will need to add a flag to consume environment yml files. 
These should be able to provide:

- url
- client id or username
- client secret or password
- whether or not to use TLS/skip-ssl-validation

We have a format that we are using already, but we could change it. Currently we're using `--auth-file`, we could do something like `--env-file`.  

Since we are consuming now from the command line, the env vars, and env file, we should decide on a hierarchy schema. 

cc @anEXPer

mcwumbly:
> We have a format that we are using already, but we could change it. Currently we're using `--auth-file`, we could do something like `--env-file`.

It's a bit redundant since it's also used as a _command_ flag in some cases, but I'm somewhat leaning toward `--config`, and opening the door to allowing any of the _options_ to be specified within the file:

```
Usage: om [options] <command> [<args>]
  --client-id, -c, OM_CLIENT_ID          string  Client ID for the Ops Manager VM (not required for unauthenticated commands)
  --client-secret, -s, OM_CLIENT_SECRET  string  Client Secret for the Ops Manager VM (not required for unauthenticated commands)
  --connect-timeout, -o                  int     timeout in seconds to make TCP connections (default: 5)
  --format, -f                           string  Format to print as (options: table,json) (default: table)
  --password, -p, OM_PASSWORD            string  admin password for the Ops Manager VM (not required for unauthenticated commands)
  --request-timeout, -r                  int     timeout in seconds for HTTP requests to Ops Manager (default: 1800)
  --skip-ssl-validation, -k              bool    skip ssl certificate validation during http requests (default: false)
  --target, -t, OM_TARGET                string  location of the Ops Manager VM
  --trace, -tr                           bool    prints HTTP requests and response payloads
  --username, -u, OM_USERNAME            string  admin username for the Ops Manager VM (not required for unauthenticated commands)
```

> Since we are consuming now from the command line, the env vars, and env file, we should decide on a hierarchy schema.

I propose the following hierarchy:

1. `--flags` have the highest precedence
2. `ENV_VARS` take precedence over `--config` values, but are overridden by `--flags`
3. `--config` values have the lowest precedence

anEXPer:
## On Naming
[...]

PR #223, opened 2018-07-31
Add env file flag support

fredwangwang:
PR for issue #204


mcwumbly:
Hey @fredwangwang just a heads up that our team is off site this week so we may be a bit slow to respond to this PR, but I'll check in with folks in the morning and share another update here afterwards.

mcwumbly:
We will make some time to look at this and catch up with you tomorrow!

PR #232, opened 2018-08-15
Added functionality to update OpsMan tempest cert

### LINKED ISSUE #180: Feature Request: Updating Opsman tempest cert
Would a PR allowing the om-cli to update the tempest cert on the Opsman VM be considered? The current process is outlined here [in the documentation](https://community.pivotal.io/s/article/How-to-Replace-Self-Signed-Certificate-in-Operations-Manager), however the process involves manually replacing a file via SSH. This process could be automated and added to the om-cli to SSH in and replace the cert using either a SSH-User and Pass or SSH-PrivateKey. This would allow [pcf-pipelines](https://github.com/pivotal-cf/pcf-pipelines) or other automation to add the correct certificate and cut down on mistakes and time when deploying a new environment.

daviddob:
Adds three commands and closes #180 

- ssl-certificate (Gets current applied tempest cert)
- update-ssl-certificate (Updates the tempest cert to one supplied by the user)
- delete-ssl-certificate (Deleted the custom cert and reverts to pre-generated tempest cert)


jtarchie:
Are you able to proive an example workflow how how (and why) to use these commands?

We'd like to verify they are working as expected not just from the CLI level, but how they affect Ops Manager.

daviddob:
The why can be summed up with "I dont want my OpsManager to be running with a self-signed certificate and the warnings associated with that (i.e. having to use -k everywhere and dealing with the browser exceptions)"

The how can be summed up as follows:
```
./om -k ssl-certificate
+-------------------------------+
|         CERTIFICATE           |
+-------------------------------+
| Ops Manager Self Signed Cert  |
+-------------------------------+
```
^ Gets the currently installed user provided ssl-cert (if applicable) otherwise defaults to the above
```
./om -k update-ssl-certificate --certificate-pem "$(safe get secret/omcert:certificate)" --private-key-pem "$(safe get secret/omcert:key)"
Successfully applied custom SSL Certificate.
Please allow about 1 min for the new certificate to take effect.
```
^ Sets the certificate on opsman to the one specified by `--certificate-pem` and `--private-key-pem`
In this case I am using safe to pull the cert out of a hashicorp vault, however those could easily be replaced with a raw string of the cert itself, a cat from a cert file, etc.

```
./om ssl-certificate
+------------------------------------------------------------------+
|                           CERTIFICATE                            |
+------------------------------------------------------------------+
| -----BEGIN CERTIFICATE-----                                      |
| MIIE4TCCAsmgAwIBAgIBAjANBgkqhkiG9w0BAQ0FADAYMRYwFAYDVQQDEw1wY2Yy |
| Mi5zbncuY29tMB4XDTE4MDgxNTE2NTYwOVoXDTI4MDgxMjE2NTYwOVowGDEWMBQG |
|                                ...                               |
| QQI/wxn6bu00hJpQ9ouDaLl2seV/pGwJS89f9FTkQ3wK2VirhCan0JCw2uBnb13y |
| s2uo6jLYYTNcWUo7y14ZJa+4BAd9OnDC8ZjmaAjn75xZPmne4O5fmx+w+iOYRU3E |
| VnQawrk=                                                         |
| -----END CERTIFICATE-----                                        |
+------------------------------------------------------------------+
```
This change is reflected in the call to `./om ssl-certificate` as show above.

At this point, opsman has changed over to using the new certificate and provided it is a valid signed cert (or a user trusted certificate) then they are able to visit opsmanager in the browser without warnings/browser exceptions, om cli without -k, etc.

If the operator wishes to revert back to the self-signed certificate, they can use `./om delete-ssl-certificate` as shown below.

```
./om delete-ssl-certificate
Successfully deleted custom SSL Certificate and reverted to the provided self-signed SSL certificate.
Please allow about 1 min for the new certificate to take effect.
```

api/ssl_certificates_service.go
```diff
@@ -0,0 +1,92 @@
...
+		return SSLCertificateOutput{}, err
+	}
+
+	err = json.NewDecoder(resp.Body).Decode(&output)
+	if err != nil {
+		return output, err
+	}
+
+	if output.Certificate.Certificate == "" {
+		output.Certificate.Certificate = "Ops Manager Self Signed Cert"
```
fredwangwang: This definitely makes the output nicer when using the table format. But when using json, it is hard to be consumed programmatically, for two reasons:
1. the string is not actually valid cert, accidentally parsing it into x509 cert would cause an error if not caught.
1. it introduces a potential fragile dependency for the consumer. If the exact string match is used by the consumer, they need to rewrite the logic each time the hardcoded string changes (though not likely to change).

So what about moving the string out from `api` package? `GetSSLCertificate()` should respect the payload returned from opsman. And if we want a nice table output, we can have the hardcoded string added to the table at `ssl-certificate` command level. 

This way we can have the raw json if we use `-f json`, or a nice table if we use `-f table`.

fredwangwang:
@daviddob Thanks for the PR! The only question I have about this PR is the hardcoded string "Ops Manager Self Signed Cert". I left a review there. Would you like to change/ discuss it? Otherwise it looks good to me!

@mcwumbly Do you have any opinion regarding the name of the commands?

api/ssl_certificates_service.go
daviddob: Thats a good suggestion, the raw json happens to be empty string in the case of an OpsMan self signed certificate. Would that be preferable for the json format and have a `if table then add Self signed cert string` case? In terms of case 2 I also agree, however I doubt it would need to change in the future, and if they were using the string match in a script it would be via the json format (I hope) and end up with an empty string which should be easier to check for than a string. Thoughts?

fredwangwang: actually the raw json should be `null` according to the opsman api docs, see [here](http://docs.pivotal.io/pivotalcf/2-3/opsman-api/#custom-ssl-certificate-settings). The empty string you got is because the type you used (string). If `interface{}` is used, then you would get the raw representation.

And for the second point, I would agree that it is not likely to change, but I would prefer to do it "right" if we can instead of assumption bad thing would not happen.

And to the question empty string, it depends on you. If you'd like to change it to interface, it would be an issue. But even empty issue would be easier to programmatically to handle than a hardcoded string. It is a stronger contract.

mcwumbly:
> Do you have any opinion regarding the name of the commands?

No strong opinions here. `ssl-certificate` seems consistent with the API, but I could see shortening it to `certificate` if others think `ssl-` is superfluous.

kcboyle:
@daviddob We were looking at this PR again today, and would love to pull it in, but are really unfamiliar with the feature. Are you available for a zoom/appear.in/messenger/etc to talk about it?

fredwangwang:
manually merged https://github.com/pivotal-cf/om/commit/8fb40aabe8dbf751b60bf9591a7ec2ef28fd038f

PR #242, opened 2018-08-31
Add retries on networking errors and 5XX responses

ljfranklin:
- `om` will now retry http requests twice in the event of networking errors or
  500 response codes


ljfranklin:
We've been intermittently seeing the following error in our CI:

```
processing product
beginning product upload to Ops Manager
 10.64 GiB / 12.60 GiB [==================================>------]  84.44% 1m34s
could not execute "upload-product": failed to upload product: could not make api request to available_products endpoint: Post https://pcf-optional.abalone-scale.gcp.releng.cf-app.com/api/v0/available_products: EOF
```

I'm hoping adding retries is a potential fix.

Issue #246, opened 2018-09-05
Add Configure-authentication config file support


fredwangwang:
Story to unwrap [`configure-authentication`](https://www.pivotaltracker.com/story/show/159974379)

Currently, we have a wrapper around `configure-authentication` to allow the configurations to be passed in using `--config config.yml`. Would like to merge that behavior into `om`.

Proposed changes:
adding a `--config` flag which takes configuration file for auth.

File format: replicate flat flag structure
example:
```yaml
username: 
password: 
decryption-passphrase: 
http-proxy-url: 
https-proxy-url: 
no-proxy: 
```

fredwangwang:
Closing this, merged.

PR #252, opened 2018-09-06
add config file support for configure-authentication

fredwangwang:
Re #246 

Now `configure-authentication` takes a `--config config.yml`.

The example format:
```yaml
username: "user"             
password: "pass"             
decryption-passphrase: "dp"
http-proxy-url: "http://proxy.myenterprise.com"       
https-proxy-url: "https://proxy.myenterprise.com"      
no-proxy: "127.0.0.1,10.0.0.2"             
```


acceptance/help_test.go
```diff
@@ -92,12 +92,13 @@ Usage: om [options] configure-authentication [<args>]
   --version, -v                          bool    prints the om release version (default: false)
 
 Command Arguments:
-  --decryption-passphrase, -dp  string (required)  passphrase used to encrypt the installation
-  --http-proxy-url              string             proxy for outbound HTTP network traffic
-  --https-proxy-url             string             proxy for outbound HTTPS network traffic
-  --no-proxy                    string             comma-separated list of hosts that do not go through the proxy
-  --password, -p, OM_PASSWORD   string (required)  admin password
-  --username, -u, OM_USERNAME   string (required)  admin username
+  --config, -c                  string  path to yml file containing authentication configuration
```
ljfranklin: The previous help text described what was required and what was optional, I think we need to bring that back in some fashion. Maybe re-add `(required)` next to username and password and change the `config` description to say that it's an alternative to passing the other flags and the config file should contain the flag names as top-level keys.

fredwangwang: 👌

commands/configure_authentication.go
```diff
@@ -18,12 +21,13 @@ type ConfigureAuthentication struct {
...
-		Password             string `long:"password"              short:"p"  env:"OM_PASSWORD"  required:"true" description:"admin password"`
-		DecryptionPassphrase string `long:"decryption-passphrase" short:"dp"                    required:"true" description:"passphrase used to encrypt the installation"`
-		HTTPProxyURL         string `long:"http-proxy-url"                                                      description:"proxy for outbound HTTP network traffic"`
-		HTTPSProxyURL        string `long:"https-proxy-url"                                                     description:"proxy for outbound HTTPS network traffic"`
-		NoProxy              string `long:"no-proxy"                                                            description:"comma-separated list of hosts that do not go through the proxy"`
+		ConfigFile           string `                             long:"config"                short:"c"                    description:"path to yml file containing authentication configuration"`
+		Username             string `yaml:"username"              long:"username"              short:"u"  env:"OM_USERNAME" description:"admin username"`
```
ljfranklin: Since we're already diving into the reflection rabbit hole, can we do something fancy to avoid needing to copy the `long` struct tag into the `yaml` tag? Guessing reflection gives you a way to programmatically add struct tags, might be possible to add the `yaml` tag on the fly before passing to yaml parser.

ljfranklin:
Left a couple comments but overall it's looking good to me 👍

commands/configure_authentication.go
fredwangwang: the `reflect.tag` only supports read, not write unfortunately. 
An alternative way is to unmarshal the given config file into a `map[string]interface{}`, then inspect what `long` tag the field has and back-fill the corresponding map item into the struct.

It essentially achieves the effect of eliminating `yaml` tag, but means we put a little bit more unmarshal logic into the function instead of in the `yaml.unmarshal`

PR #253, opened 2018-09-06
Adds --vars-env flag

acrmp:
This adds a `--vars-env flag` similar to the BOSH cli.

One difference in behaviour is that the BOSH cli does not handle multi-line values for environment variables (the value is treated as YAML and newlines are converted to spaces). In this version we treat strings specially to avoid losing newlines.

Our original motivation for this was to add support for environment variables to the `configure-product` command, but we've also updated the `configure-director`, `create-vm-extension` and `interpolate` commands.

The help output included in the docs directory was stale so we've updated that as well.

Thanks,

@acrmp and @michelleheh (CC @ljfranklin)


jtarchie:
This looks good to us! (@fredwangwang and I)

We tested it out like:

```bash
$ export SOME_SSH_KEY="$(cat ~/.ssh/id_rsa)"
$ go run main.go interpolate -c <(echo "key: ((SSH_KEY))") --vars-env SOME
key: |-
  -----BEGIN RSA PRIVATE KEY-----
....
  -----END RSA PRIVATE KEY-----
$ go run main.go interpolate -c <(echo "key: ((SSH_KEY123))") --vars-env SOME
could not execute "interpolate": Expected to find variables: SSH_KEY123
exit status 1
```

The only question we had was respect to the `envFunc`. Was this done purely for testing?

PR #260, opened 2018-09-19
Adding new flag for Apply Changes

hjaffan:
To reduce the amount of time it takes to apply changes for things that have not changed. This pull request adds a flag to apply changes to skip unchanged products.

This will get the list of pending changes and populate them into deployProducts variable. 

If there are no pending changes, we default to not deploying any apps.


pivotal-issuemaster:
@hjaffan Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/pivotal?repositoryId=pivotal-cf/om&pullRequestId=260)!

[Click here](https://cla.pivotal.io/sync/pivotal?repositoryId=pivotal-cf/om&pullRequestId=260) to manually synchronize the status of this Pull Request.

See the [FAQ](https://cla.pivotal.io/about) for frequently asked questions.

jtarchie:
Are you able to proive an example workflow how how (and why) to use these commands? We'd like to verify they are working as expected not just from the CLI level, but how they affect Ops Manager.

main.go
```diff
@@ -135,7 +135,7 @@ func main() {
 
 	commandSet := jhanda.CommandSet{}
 	commandSet["activate-certificate-authority"] = commands.NewActivateCertificateAuthority(api, stdout)
-	commandSet["apply-changes"] = commands.NewApplyChanges(api, logWriter, stdout, applySleepSeconds)
+	commandSet["apply-changes"] = commands.NewApplyChanges(api, api, logWriter, stdout, applySleepSeconds)
```
fredwangwang: is there a particular reasoning behind using two api service instead of extending `applyChangesService` interface to include `ListStagedPendingChanges()` method?

pivotal-issuemaster:
@hjaffan Thank you for signing the [Contributor License Agreement](https://cla.pivotal.io/sign/pivotal?repositoryId=pivotal-cf/om&pullRequestId=260)!

hjaffan:
@jtarchie So the main reason behind this additional flag is to only deploy actual changes.
The workflow we are working with is for the PCF Deployment pipeline where we have all tiles being configured prior to applying changes, and then say we add a new tile that did not alter the configurations of the other existing tiles, rather than applying changes to everything all the time, this additional flag will just run the apply changes for that specific product.

main.go
hjaffan: Hi @fredwangwang there was no specific reason, we were just trying to do it in a way that would be least intrusive if that answers your question?

jtarchie:
@hjaffan, I understand the usage of want. I'm looking for steps to verify.

Can you provide accurate instructions for us to verify that the features works as specified?

fredwangwang:
manually merged in this commit: https://github.com/pivotal-cf/om/commit/f87449add8c0abaf9537b3fc7ab397d6435958bb

Issue #261, opened 2018-09-21
retrieve errand for `staged-config`


fredwangwang:
`configure-product` is now able to set the errand state for the given product, but its counterpart `staged-config` could not retrive errand.
It makes sense that `staged-config` would output the errand info as well.

Format:
to match the input format of `configure-product`, the errand section should look like:
```yaml
errand-config:
  smoke_tests:
    post-deploy-state: true
    pre-delete-state: default
  push-usage-service:
    post-deploy-state: false
    pre-delete-state: when-changed
```

it is for the use case:
`staged-config` (from one foundation)-> `configure-product` (to another foundation). 

But right now `staged-config` misses `errand` props, so it is helpful that `staged-config` can return that piece of info to enable the migration of errand state.

jtarchie:
merged

PR #262, opened 2018-09-21
add retrive errand for staged-config

fredwangwang:
re issue #261 

expected output:
```
$ om staged-config -p cf
<truncated>
errand-config:
  deploy-autoscaler:
    post-deploy-state: true
  deploy-notifications:
    post-deploy-state: false
  deploy-notifications-ui:
    post-deploy-state: false
  nfsbrokerpush:
    post-deploy-state: false
  push-apps-manager:
    post-deploy-state: false
  push-usage-service:
    post-deploy-state: false
  smoke_tests:
    post-deploy-state: true
  test-autoscaling:
    post-deploy-state: false
```

### DIFF
--- a/acceptance/staged_config_test.go
+++ b/acceptance/staged_config_test.go
@@ -115,6 +115,18 @@ var _ = Describe("staged-config command", func() {
 							}
 						}
 					}`))
+			case "/api/v0/staged/products/some-product-guid/errands":
+				w.Write([]byte(`{ "errands": [
+                           {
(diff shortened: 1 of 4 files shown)


jtarchie:
Since the APIs of OpsManager only returns the `true` or `false`, you'll never get the `default` string back. The `default` string is something that can be [set via the API endpoint](https://docs.pivotal.io/pivotalcf/2-3/opsman-api/#configuring-errands), though.

Issue #221, opened 2018-07-26
Remove DEPRECATED commands


heycait:
We have a some commands marked as \*\*DEPRECATED**.

Currently, `configure-bosh` is deprecated in favor of `configure-director` and `set-errand-state` with `configure-product`. Probably best to remove these before any official `om` CLI comes out.

jtarchie:
Closing as the two PRs were merged in.

PR #263, opened 2018-09-26
Remove configure-bosh command

mcwumbly:
The `configure-bosh` command has been deprecated for some time in favor
or the `configure-director` command.

[pivotal-cf/om#221]


jtarchie:
This would break usage in [pcf-pipelines](https://github.com/pivotal-cf/pcf-pipelines/blob/master/install-pcf/openstack/tasks/config-director/task.sh#L190) at the moment. We are investigating what it would take to remove this.

PR #264, opened 2018-09-26
Remove deprecated set-errand-state command

mcwumbly:
[pivotal-cf/om#221]


jtarchie:
This would break usage in [pcf-pipelines](https://github.com/pivotal-cf/pcf-pipelines/blob/master/tasks/toggle-errands/task.sh#L46) at the moment. We are investigating what it would take to remove this.

We've also validated that third party users are using it, too.

mcwumbly:
one possible option for folks is to pin back to a version of `om` that works until they can consume the breaking change.

jtarchie:
We are locking at a current version of om on pcf-pipelines. Merging.

In commit 0dbd0f63adc68a487aa77e20072c61cf4cdea61f, rebased on master.

PR #272, opened 2018-10-05
Add config file support to upload product

### LINKED ISSUE #270: add config file support for upload-product
related to our unwrapping story: https://www.pivotaltracker.com/story/show/160566798

currently `p-automator upload-product` allows a flag `-c` to accept a file containing optional shasum & version validation. 
There are existing patterns in `om` that we can supply any supported commandline flag through the config file, we would like to extend that behavior to `upload-product` as well.

Sample config file:
```yaml
shasum:
version: 
polling-interval:
product:
```

@mcwumbly Also, I want to discuss the flag name `shasum`. It does not indicate which sha version we are using, probably rename it to sha2 makes more sense?

fredwangwang:
resolves #270


anEXPer:
In the course of in-person review, we determined that the `--version` flag this adds may be confusing, and should change to `--product-version`. However, we decided to stop short of putting `product-` in front of all the other flags, as that would stack up unnecessary breaking changes for no particular benefit. We'll make these changes presently.

PR #286, opened 2018-11-14
prevent stage/configure product/director if apply-change is running

fredwangwang:
If a product is staged/configured while apply-change is running,
the product will be unstaged/unconfigured after the apply-change
finishes. This is an unexpected behavior to the users, so we want
to prevent stage/configure product/director if there is a running
installation.

The change in this commit will fail fast if the cli user tries to
stage/configure if apply-change is running.

The prevention logic will eventually live in Ops Manager as well.

[#161630968]

### DIFF
--- a/acceptance/configure_director_test.go
+++ b/acceptance/configure_director_test.go
@@ -42,6 +42,8 @@ var _ = Describe("configure-director command", func() {
 			w.Header().Set("Content-Type", "application/json")
 
 			switch req.URL.Path {
+			case "/api/v0/installations":
+				w.Write([]byte(`{"installations": []}`))
 			case "/uaa/oauth/token":
 				username := req.FormValue("username")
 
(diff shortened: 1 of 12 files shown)


ljfranklin:
@fredwangwang sorry for the delay in reviewing this, most of our team was out last week. I also would not have expected OpsMgr to react that way if you ran `configure-product` while an apply changes was running. Did you report the issue to OpsMgr? Sounds like a bug to me. But in the meantime this looks like a nice guardrail for `om` users. Feel free to merge this in whenever 👍

fredwangwang:
@ljfranklin thanks for reviewing! We've talked with OpsManager team, they will introduce the guard probably in 2.4-2.5. We are adding the logic now in `om` side as a workaround to simulate the future behavior of OpsManager

Issue #290, opened 2018-11-19
Feature Request: eval "$(om bosh-env)"


genevieve:
Hey there! 

I'm working on the setup for a test suite that reminds me of `eval "$(bbl print-env)"` but `eval "$(om bosh-env)"`. 

It would curl for the bosh command line creds, get the bosh ca cert, create a local tmp file for the cert to point BOSH_CA_CERT at, and then export BOSH_ALL_PROXY. It could also just create a config file that you pass to the bosh cli with the `--config` file if it supports bosh_all_proxy.

Is this something you would consider adding to the om cli? 

```go
		// Get $BOSH_CLIENT, $BOSH_CLIENT_SECRET, $BOSH_ENVIRONMENT
		output, err := om.Curl(api.RequestServiceCurlInput{
			Path:   "/api/v0/deployed/director/credentials/bosh_commandline_credentials",
			Method: "GET",
		})
		Expect(err).NotTo(HaveOccurred())

		// Decode the response body
		var creds CredentialOutput
		err = json.NewDecoder(output.Body).Decode(&creds)
		Expect(err).NotTo(HaveOccurred())

		// Export the bosh commandline credentials
		Expect(creds.Credential).To(ContainSubstring("BOSH_ENVIRONMENT=10.0.0.5 bosh"))
		kvs := strings.Split(creds.Credential, " ")
		for _, kv := range kvs {
			if strings.Contains(kv, "=") {
				k := strings.Split(kv, "=")[0]
				v := strings.Split(kv, "=")[1]
				os.Setenv(k, v)
			}
		}

		// Get the bosh ca cert
		output, err = om.Curl(api.RequestServiceCurlInput{Path: "/api/v0/certificate_authorities", Method: "GET"})
		Expect(err).NotTo(HaveOccurred())

		// Decode the response body
		var cas CAOutput
		err = json.NewDecoder(output.Body).Decode(&cas)
		Expect(err).NotTo(HaveOccurred())
		certPem := cas.CertificateAuthorities[0].CertPem

		// Create a tmp file for the bosh ca cert
		caFile, err := ioutil.TempFile("", "bosh-ca-cert")
		Expect(err).NotTo(HaveOccurred())
		_, err = caFile.WriteString(certPem)
		Expect(err).NotTo(HaveOccurred())
		ca := caFile.Name()

		// Export BOSH_ALL_PROXY
[...]

PR #292, opened 2018-11-21
Adds support similar to bbl print-env

calebwashburn:
Adds feature referenced in issue (https://github.com/pivotal-cf/om/issues/290)


mcwumbly:
That was quick! ...though this feels like a large dependency to be pulling in for this feature. 

I'm thinking "[a little copying is better than a little dependency](https://go-proverbs.github.io/)" in this case.

calebwashburn:
@mcwumbly I struggled with to vendor or copy the renederer package.  Happy to refactor to copy and paste or borrow the logic. Liked the idea that this supported both postfix and powershell like bbl.  Let me know how we want to proceed.

mcwumbly:
nice! 

it may be worth squashing these commits before merging so that the blobs from the add/remove bbl isn't carried around in git history forever

calebwashburn:
@genevieve Sorry didn't see your note until I had already removed the `bbl` dependency

jtarchie:
Just checked this out. It worked™!

Only thing we noticed. If the `target` is specified with a protocol schema then that gets used in the full address for the SOCKS5 proxy. ie `ssh+socks5://ubuntu@https:/example.com22?private-key=ssh-key`.

calebwashburn:
@jtarchie Added logic to remove protocol scheme if present.

PR #298, opened 2018-12-04
Retry when there's an EOF error while uploading tile

### LINKED ISSUE #240: Intermittent `POST .../api/v0/available_products: EOF` when uploading tile
We've seen the following error 22 times in the last 45 days:

```
processing product
beginning product upload to Ops Manager
 10.64 GiB / 12.60 GiB [==================================>------]  84.44% 1m34s
could not execute "upload-product": failed to upload product: could not make api request to available_products endpoint: Post https://pcf-optional.abalone-scale.gcp.releng.cf-app.com/api/v0/available_products: EOF
```

We've seen this across multiple tiles (PAS & PASW) and different IaaS (GCP, Azure, AWS) and different OpsMgr versions (2.0-2.3). We initially suspected a recent [PR](https://github.com/pivotal-cf/om/pull/211) which made file uploads more performant but we saw the error a few times prior to merging the PR.

One potential solution would be to retry on `Temporary()` networking errors. The stdlib networking packages often return `net.Error` which has a `Temporary()` method you can use to check whether an error instance might be retryable: https://golang.org/pkg/net/#Error. We could add retry logic for these errors in the http client in `om`: https://github.com/pivotal-cf/om/blob/4d5f262bb6a1006f1e2af2754ee4e24707b5e4f3/network/unauthenticated_client.go. We suspect `EOF` is a temporary error but we're not positive.

genevieve:
Fixes #240

cc @ljfranklin


jtarchie:
@genevieve and @ljfranklin, I thought that this had been solved already a few months back. Are you all experiencing it again? If so, does this resolve the error for you in your pipeline?

ljfranklin:
@jtarchie no, our initial pass a while ago didn't help resolve the issue. More context here: https://github.com/pivotal-cf/om/issues/240#issuecomment-429910526. We haven't got a chance to put any mileage on this in our CI yet, but we were able to reproduce the error with an acceptance test included in the PR.

fredwangwang:
> adding a Reset method on the formcontent struct.

I wanted to add that a while back but keeping forget... Nice work!

But meanwhile it seems still worth to investigate supproting resumable upload. Adding something like [this](https://www.nginx.com/resources/wiki/modules/upload/#upload-resumable) into OpsManager and om would probably better solve the problem than retrying uploading the complete payload several times.

ljfranklin:
@fredwangwang @jtarchie could y'all give this a look when you get a chance? I agree resumable uploads would be great but getting the OpsMgr nginx config updated and backported to all supported versions seems like a tall order at this time.

Issue #277, opened 2018-10-19
go mod


fredwangwang:
Do we want to use vgo?

patrick-huber-pivotal:
Looks like you just need to update the go version to 1.11 to take advantage of modules

https://github.com/golang/go/wiki/Modules#quick-start-example


> Go 1.11 includes preliminary support for versioned modules as proposed here. Modules are an experimental opt-in feature in Go 1.11, with the hope of incorporating feedback and finalizing the feature for Go 1.12. Even though the details may change, future releases will support modules defined using **Go 1.11** or **vgo**.

> The recent work by the Go team on versioned modules started outside of the main Go repository with the vgo tool, but on July 12, 2018 support for versioned modules landed in the main repository (announcement thread), and Go 1.11 was released on August 24, 2018.

zachgersh:
cc @jtarchie - we can close this now.

PR #309, opened 2018-12-22
Officially Move `om` over to go modules

zachgersh:
The go world is changing and we are looking at a whole new world of dependency management (go modules) built directly into the go tool chain.

I've switched `om` over and in addition removed the `vendor` directory. This means you will need to tweak your build pipeline in one of two pretty simple ways:

1) if inside of the `GOPATH`: `GO111MODULE=on go build .`
2) if you want to build the project and not setup a `GOPATH`: `go build .`


zachgersh:
Also, you are totally going to want to just look at the second commit and not the removes Gopkg commit - that thing is just removing a bunch of files :/

jtarchie:
We might not be able to change yet. Will discuss via slack about the OSL process.

jtarchie:
Have confirmed we can do this by OSL team.

To do list before fully supporting.

* [ ] need to migrate CI tasks to use `go mod` instead of `dep`
* [ ] local workflow changes, as we no longer use `GO_PATH` for this in development
* [ ] how does this affect our upstream dependency `p-automator`, which uses `dep` at the moment

zachgersh:
I am going to fix this conflicting file! @jtarchie would it work better if we merged this to a particular transition branch so you can fix CI first? I can also take a look at p-automator to see what sort of changes would be required if you like.

fredwangwang:
issue #277

jtarchie:
Manually merged in.

Updating tasks to use it correctly.

Issue #92, opened 2017-07-19
Feature Request: Add deploy-errands flag to apply-changes command


DonBower:
According to the documentation, [Triggering an install process](http://opsman-dev-api-docs.cfapps.io/#triggering-an-install-process),  we can set which errands to run during an apply changes.
This is very useful, especially when making changes to something other that ERT.

Although the implementation of the **/api/v0/installations** endpoint varies between 1.9 and 1.11, it would be nice to be able to at least say, "apply changes and run these errands too..." as most, if not all, of our post-deploy errands are disabled.

DonBower:
Here is how we may be able to implement:
We could form the json string similar to below, and the tool would substitute the product-name for the product-guid:

```
{
  "enabled_errands": {
    "<product-name, i.e. cf>": {
      "post_deploy_errands": [
        "smoke-tests",
        "push-apps-manager",
        "notifications",
        "notifications-ui",
        "push-pivotal-account",
        "autoscaling",
        "autoscaling-register-broker"
      ]
    }
  }
}
```

kenchenpcf:
@DonBower Is this a feature that is still useful?

DonBower:
We use om in pipielines to kick off apply-changes.
To be able to specify which errands to run at the same time would be beneficial, because today, we use a curl command to start the apply with the desired errands, which returns immediately, and then we call om again with the apply command to “attach” to the current apply to make the pipeline synchronous.

mcwumbly:
We spec'd this out in [this tracker story](https://www.pivotaltracker.com/n/projects/2145041/stories/155641251), but would happily accept a PR if someone wants to pick it up sooner.

---

As a om user, I want to specify the errands to run post deploy and pre delete, so that I can customize the errands when applying changes.

`om apply-changes --errands` or `om apply-changes -e`

This mirrors the functionality in the Ops Man API described here: http://opsman-dev-api-docs.cfapps.io/#triggering-an-install-process

om apply-changes --errands-config /some/errands/config.yml

/some/errands/config.yml
```
---
product1_name:
  pre-delete:
    errand_a: true
    errand_b: false
  post-deploy:
    errand_c: "default"
product2_name:
  post-deploy:
    errand_a: false
  pre-delete:
    errand_b: "when-changed"
```

The post deploy and pre delete errand states should mirror the Ops Man API options.

See related github request here: https://github.com/pivotal-cf/om/issues/92

**Acceptance:**
Given that I run `om apply-changes`,
[...]

PR #323, opened 2019-02-08
allow errand configuraiton during apply-changes 

fredwangwang:
Allows for errand configuration during apply changes via a configuration yaml file

Fixes https://github.com/pivotal-cf/om/issues/92

[https://www.pivotaltracker.com/story/show/161848001]


fredwangwang:
While doing it we also created a helper function called `fetchProductGUID()` to fetch product guid from both `staged-product` and `deployed-product`, since products that are staged for deletion are not shown up in `staged-product` endpoint.

This brings up a refactor opportunity to change all the guid fetching logic to use the helper function, since there is a edge case that `staged-product` does not contain the guid of the product. 

But we didn't spend the time doing it, as it didn't appears to a problem in the past and it does not affect the functionaility we added.

kcboyle:
@fredwangwang @rheaton 
We are about to pull this in, but we've been testing it on the command line, and couldn't get the errands to skip. Is there something missing? using the format defined in the readme. 

We have the following config:
```
errands:
  cf:
    smoke_tests: false
    push-usage-service: false
    push-apps-manager: false
    deploy-notifications: false
    deploy-notifications-ui: false
    deploy-autoscaler: false
    test-autoscaling: false
    metric_registrar_smoke_test: false
    nfsbrokerpush: false
    smbbrokerpush: false
```

and run: `om --env env.yml apply-changes -c config.yml`, but we get output for each of the errands we thought we disabled. Is there a layer missing that just needs more documentation?

fredwangwang:
that errands config format is wrong. Here is an example (its in the issue as well):
```
errands:
  product1_name:
    run_pre_delete:
      errand_a: true
      errand_b: false
    run_post_deploy:
      errand_c: "default"
  product2_name:
    run_post_deploy:
      errand_a: false
    run_pre_delete:
      errand_b: "default"
```

we dicided to go with this because it's the format used in the api docs directly: http://docs.pivotal.io/pivotalcf/2-3/opsman-api/#triggering-an-install-process

kcboyle:
validated. Tested. merged.

PR #332, opened 2019-03-03
Skip subdirectories when locating metadata file

mcwumbly:
This issue was seen in the wild when a tile was created by zipping up
files on a Mac OSX system where it grabbed  a `__MACOSX` directory of hidden files.

`om` found the `__MACOSX/metadata/._metadata.yml` before it found the
`metadata/metadata.yml` and returned the error when `upload-product` was called:

```
could not execute "upload-product": failed to extract product metadata: could not extract product metadata: yaml: control characters are not allowed
```

This change to the regex changes `om` so that metadata files are only found
in `metadata/` or `./metadata`


anEXPer:
Thanks! Also nice cleanup on those tests.

Issue #335, opened 2019-03-07
om incorrectly reports an error when an ops manager verifier triggers a warning


Logiraptor:
We are using ops manager in an environment which requires special IAAS configuration. Because of this, some Ops Manager verifiers fail. We have disabled those verifiers, and they now simply report a warning in the UI, and a 207 (multi-status) in the API. `om` incorrectly reports that as an error and stops configuring the director.

Here is the error we get on `configure-director`:

```
$ om -e combine-env.yml configure-director --config director-template.yml --vars-file om-vars.yml
started configuring director options for bosh tile
finished configuring director options for bosh tile
started configuring availability zone options for bosh tile
successfully fetched AZs, continuing
could not execute "configure-director": availability zones configuration could not be applied: request failed: unexpected response:
HTTP/1.1 207 
Transfer-Encoding: chunked
Cache-Control: no-cache, no-store
Connection: keep-alive
Content-Type: application/json; charset=utf-8
Date: Thu, 07 Mar 2019 22:20:50 GMT
Expires: Fri, 01 Jan 1990 00:00:00 GMT
Pragma: no-cache
Server: Ops Manager
Strict-Transport-Security: max-age=15552000; includeSubDomains
X-Content-Type-Options: nosniff
X-Frame-Options: SAMEORIGIN
X-Request-Id: 64df4a71-9ef7-4c93-876a-38a01550426b
X-Runtime: 62.353585
X-Xss-Protection: 1; mode=block

ec
{"verifications":{"warnings":{"base":["Availability zone(s) were saved but there are verification errors which may prevent you from deploying","Could not log in: Connection failed: execution expired, type: AvailabilityZonesVerifier"]}}}
0

```

You can see that the HTTP status is 207, and the body includes a warning about the AvailabilityZonesVerifier.

I believe one way to fix this would be to change the line [here](https://github.com/pivotal-cf/om/blob/f6abe8a489aa9c85bcbbe7d2e54735eeb46fce62/api/validation.go#L12) to instead check that Status is <= 299 and >= 200. 

I'm happy to prepare a PR if that would be helpful.

kcboyle:
Thanks for opening an issue! We've been discussing this internally for a while, but wanted to give some OS visibility to some of the main points. 
[...]

PR #338, opened 2019-03-11
Add --ignore-verification-warnings flag

Logiraptor:
- Ops Manager returns HTTP 207 in some cases to signal a verification warning
- This is different from a verification error, which is a 422

Relevant issue #335

### DIFF
--- a/acceptance/configure_director_test.go
+++ b/acceptance/configure_director_test.go
@@ -29,6 +29,7 @@ var _ = Describe("configure-director command", func() {
 		networksPutCallCount         int
 		resourceConfigMethod         string
 		resourceConfigBody           []byte
+		verifierErrorOccured         bool
 
 		server *httptest.Server
 	)
@@ -38,6 +39,7 @@ var _ = Describe("configure-director command", func() {
 		directorPropertiesCallCount = 0
 		networkAZCallCount = 0
 		networksPutCallCount = 0
+		verifierErrorOccured = false
 
 		server = httptest.NewTLSServer(http.HandlerFunc(func(w http.ResponseWriter, req *http.Request) {
 			w.Header().Set("Content-Type", "application/json")
@@ -83,6 +85,9 @@ var _ = Describe("configure-director command", func() {
 
 					azPutCallCount++
 
+					if verifierErrorOccured {
+						w.WriteHeader(207)
+					}
 					_, err = w.Write([]byte(`{}`))
 					Expect(err).ToNot(HaveOccurred())
 				} else {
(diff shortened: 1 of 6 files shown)


pivotal-issuemaster:
@Logiraptor Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/pivotal?repositoryId=pivotal-cf/om&pullRequestId=338)!

[Click here](https://cla.pivotal.io/sync/pivotal?repositoryId=pivotal-cf/om&pullRequestId=338) to manually synchronize the status of this Pull Request.

See the [FAQ](https://cla.pivotal.io/about) for frequently asked questions.

Logiraptor:
Closing to fix CLA issue

Logiraptor:
Closing to fix CLA issue (this time with pivotal-cf org)

pivotal-issuemaster:
@Logiraptor Thank you for signing the [Contributor License Agreement](https://cla.pivotal.io/sign/pivotal?repositoryId=pivotal-cf/om&pullRequestId=338)!

acceptance/configure_director_test.go
```diff
@@ -543,5 +548,43 @@ properties-configuration:
 			"elb_names": ["my-elb"]
 	  }`))
 		})
+
+		It("configures the BOSH director using the API", func() {
```
kcboyle: What is unique about this test? You are specifically testing that this returns a 207, but I don't get this at all from your description

acceptance/help_test.go
```diff
@@ -84,19 +85,20 @@ const CONFIGURE_AUTHENTICATION_USAGE = `ॐ  configure-authentication
...
-  --version, -v                                          bool    prints the om release version (default: false)
+  --client-id, -c, OM_CLIENT_ID                            string  Client ID for the Ops Manager VM (not required for unauthenticated commands)
+  --client-secret, -s, OM_CLIENT_SECRET                    string  Client Secret for the Ops Manager VM (not required for unauthenticated commands)
+  --connect-timeout, -o, OM_CONNECT_TIMEOUT                int     timeout in seconds to make TCP connections (default: 10)
+  --decryption-passphrase, -d, OM_DECRYPTION_PASSPHRASE    string  Passphrase to decrypt the installation if the Ops Manager VM has been rebooted (optional for most commands)
+  --env, -e                                                string  env file with login credentials
+  --help, -h                                               bool    prints this usage information (default: false)
+  --ignore-verifier-warnings, OM_IGNORE_VERIFIER_WARNINGS  bool    continues working even if there are verifier warnings present (default: false)
```
kcboyle: why is this a global option? `--ignore-verifier-warnings` should only be enabled for `configure-product` and `configure-director`.  A good place for this that would not affect other commands would be in the `Options` structs in `ConfigureDirector` and `ConfigureProduct`

main.go
```diff
@@ -35,19 +35,20 @@ type httpClient interface {
...
-	Trace                bool   `yaml:"trace"                 short:"tr" long:"trace"               env:"OM_TRACE"                               description:"prints HTTP requests and response payloads"`
-	Username             string `yaml:"username"              short:"u"  long:"username"            env:"OM_USERNAME"                            description:"admin username for the Ops Manager VM (not required for unauthenticated commands)"`
-	Env                  string `                             short:"e"  long:"env"                                                              description:"env file with login credentials"`
-	Version              bool   `                             short:"v"  long:"version"                                          default:"false" description:"prints the om release version"`
+	DecryptionPassphrase   string `yaml:"decryption-passphrase"    short:"d"  long:"decryption-passphrase"    env:"OM_DECRYPTION_PASSPHRASE"                    description:"Passphrase to decrypt the installation if the Ops Manager VM has been rebooted (optional for most commands)"`
```
kcboyle: changes that should only affect 2 commands should not change main.go

acceptance/help_test.go
```diff
@@ -13,19 +13,20 @@ const GLOBAL_USAGE = `ॐ
...
```
kcboyle: Can we make this description something more like: 
```
 option to ignore verifier warnings. NOT RECOMMENDED UNLESS DISABLED IN OPS MANAGER (default: false) 
```

The reason for this is that we do not want users to use this flag unless they absolutely know what they are doing. We want to protect people from themselves.

acceptance/configure_director_test.go
Logiraptor: Agreed, that's an oversight on my part.

acceptance/help_test.go
Logiraptor: Sounds good.

acceptance/help_test.go
Logiraptor: I made the change to scope this flag to just configure-director, since that is the only known place where a 207 is returned.

Logiraptor:
@kcboyle Thanks for the review! I think I've addressed the feedback. 

The flag is now scoped very tightly to just the configure-director command, and internally only the call to the availability zones api. 

I believe this version should be good to go. Let me know if you have any other concerns.

Logiraptor:
Also, I notice the ci job is failing, but I don't have access to those logs. Let me know if there's something I can fix. All the tests pass for me locally.

kcboyle:
Hi @Logiraptor ! I believe the failure is an issue on our part with active development (we have pending tests, and this is not allowed). So if the tests pass locally, don't worry about it. We'll see if we can take another look today/tomorrow. Thanks for responding so quickly (and for the PR!)

kcboyle:
Merged via command line. Thanks for the PR!

PR #342, opened 2019-04-11
Fix Markdown formatting

nmahoney-pivotal:
One of the command links was broken.


kcboyle:
Good catch @nmahoney-pivotal 

Pulling this in now.

PR #353, opened 2019-04-26
delete installation now requires force flag

### LINKED ISSUE #341: delete-installation should have a confirm step
We've received feedback that someone accidentally deleted their installation, and desired a safety. 

Proposal to include an optional `--force` flag to the command. This would be a breaking change, as we would be making the automatic delete an opt-in feature, rather than the default. However, this would add a "safety" to the command if accidentally run.

zachgersh:
Let me know if the design around this does not work! I think the aesthetics are ok.

Gentle reminder, as I always forget this, change is backwards incompatible and requires a major revision of `om` when released (as we've changed the way the delete-installation command works by default).

fixes #341

### DIFF
--- a/acceptance/delete_installation_test.go
+++ b/acceptance/delete_installation_test.go
@@ -82,7 +82,8 @@ var _ = Describe("delete-installation command", func() {
 			"--username", "some-username",
 			"--password", "some-password",
 			"--skip-ssl-validation",
-			"delete-installation")
+			"delete-installation",
+			"--force")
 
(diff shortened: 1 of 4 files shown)


zachgersh:
cc @jtarchie @kcboyle 👯‍♂

kcboyle:
Hi, @zachgersh ! Thank you for the PR! We are working on a couple 2.6 support items currently, but have this PR scheduled for shortly after that work is done (about 5 stories), just to give you a general time frame around when we will be able to pull it in.

jpmcb:
@zachgersh merged! Thanks so much for the contribution.

Issue #366, opened 2019-05-17
om upload-product can show a failure if Ops Manager takes too long to respond 


pivotal-todd-ritchie:
`om` can show a failure when uploading large tiles that take a long time to decompress. As an example, I'm having trouble uploading SRT on a brand new terraformed azure opsman.

When I upload the tile, ops manager takes little over 6 minutes to decompress and validate the tile. By default, the idle connection timeout on Azure is 4 minutes. 

The result looks like this:

```
+ om -k upload-product --product ./srt-2.5.2-build.7.pivotal
processing product
beginning product upload to Ops Manager
 10.12 GiB / 10.12 GiB  100.00% 8m9s1s2s
21m50s elapsed, waiting for response from Ops Manager...
retrying product upload after error: could not make api request to available_products endpoint: Post https://pcf.example.com/api/v0/available_products: net/http: request canceled (Client.Timeout exceeded while awaiting headers)
processing product
beginning product upload to Ops Manager
...
```

If I browse to the ops manager about 10 minutes into the "waiting for response from Ops Manager" I can see the tile is available to stage. `om` sees the timeout, retries twice, and then exits with a failure after exhausting retries.

My workaround is to increase the idle connection timeout on the Azure VM. Having `om` check if the tile is there already before each retry attempt could also help.

kcboyle:
The PR has been prioritized for review.
Steps our team will go through before/after pulling it in:
- include a concrete procedure for replicating the error
- affirm that we do not error if the product is already there

To our team's understanding, this is most easily replicate-able on Azure, and is due to a server timeout on the Azure side, _not_ a timeout from `om`.

anEXPer:
We're pretty sure this isn't governed by the Azure idle timeout, but by our default `request-timeout`.

The upshot is: extend your request-timeout in your env.yml file and you won't see this problem. We've heard that "over 9000" is sufficient, but probably excessive.

We've also updated the related PR with more information about changes we'd like to see in the messaging, but increasing the timeout is probably the best play for most people, as those changes may not land for a while.

anEXPer:
Here's our followup story: https://www.pivotaltracker.com/story/show/166833456

kcboyle:
@pivotal-todd-ritchie this should have been released with better error messaging in v3.0.0. 

Closing this issue.

PR #367, opened 2019-05-17
Move the upload-product 'available versions' check into the retry loop

pivotal-todd-ritchie:
Covers the case where the first upload succeeds but times out on validation. Previously it would retry and fail. Now if the product validated, the retry can be skipped.

Co-authored-by: Pete Wall <pwall@pivotal.io>

### DIFF
--- a/commands/upload_product.go
+++ b/commands/upload_product.go
@@ -87,17 +87,18 @@ func (up UploadProduct) Execute(args []string) error {
 		up.logger.Printf("expected version matches product version.")
 	}
 
-	prodAvailable, err := up.service.CheckProductAvailability(metadata.Name, metadata.Version)
-	if err != nil {
-		return fmt.Errorf("failed to check product availability: %s", err)
-	}
-
-	if prodAvailable {
-		up.logger.Printf("product %s %s is already uploaded, nothing to be done.", metadata.Name, metadata.Version)
-		return nil
-	}
-
 	for i := 0; i <= maxProductUploadRetries; i++ {
+		var prodAvailable bool
+		prodAvailable, err = up.service.CheckProductAvailability(metadata.Name, metadata.Version)
+		if err != nil {
+			return fmt.Errorf("failed to check product availability: %s", err)
+		}
+
+		if prodAvailable {
+			up.logger.Printf("product %s %s is already uploaded, nothing to be done.", metadata.Name, metadata.Version)
+			return nil
+		}
+
 		up.logger.Printf("processing product")
 
 		err = up.multipart.AddFile("product[file]", up.Options.Product)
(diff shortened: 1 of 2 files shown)


kcboyle:
@pivotal-todd-ritchie This is a PR for issue #366, correct?

kcboyle:
The PR has been prioritized for review.
Steps our team will go through before/after pulling it in:
- include a concrete procedure for replicating the error
- affirm that we do not error if the product is already there

To our team's understanding, this is most easily replicate-able on Azure, and is due to a server timeout on the Azure side, _not_ a timeout from `om`.

pivotal-todd-ritchie:
@kcboyle Yes, that's correct. I'm happy to pair on replicating the issue on Azure.

anEXPer:
Hey @pivotal-todd-ritchie, we tried replicating this on our own Azure environment and couldn't; it works find both with and without the fix. Presumably this is because the timeout is fifteen minutes, and our unpacks are finishing in about ten.

We're having a very hard time making the unpack take more than ten minutes. How did you get it to take so long?

nbconklin:
@pivotal-todd-ritchie  friendly follow up on this, thanks!

pivotal-todd-ritchie:
Sorry for the delay, I didn't catch the earlier notification!

The terraforming-azure master branch has a patch in to extend the idle connection timeout from 4 minutes to 30 minutes. `v0.40.0` and below does not have that patch, so it ought to have a default of 4 minutes, and the 6-10 minute unpack ought to fail.

The easiest way to repro is to crank down the idle timeout in the azure portal to 4 minutes.

anEXPer:
The reason we were unable to reproduce this appears to be because we were setting the `--request-timeout` option sufficiently high. The default value for this is 1800, which is 30 minutes. We were setting it to 9001 for Dragonball reasons, so we were never seeing the problem. If you look back at the total time reported in the original issue, the request timeout happened after 30 total minutes.

The upshot is: extend your `request-timeout` in your `env.yml` file and you won't see this problem. We've heard that "over 9000" is sufficient, but probably excessive.

It still makes sense to add a check to see if the product is available before performing a retry. However, the message if the product is available on retry should be "upload succeeded", while the message for if the product is present when the command is first run should be the current "nothing to do" message. After all, the upload has in fact succeeded.

We're going to put a followup story in the backlog on this, and update/close the original issue with the `connection-timeout` workaround. If you want to update the PR to separate the messaging in the initial vs retry check before we get to it, we'd happily take that - it might be a while for us.

anEXPer:
Here's our followup story: https://www.pivotaltracker.com/story/show/166833456

nbconklin:
We're going to close this per the information and tracker story that can be followed. Please open an issue if you have any additional questions/issues, thank you!

johncornish:
We re-opened and merged this pursuant to work on our follow-up story.

Thanks again for the issue and the PR!

Issue #360, opened 2019-05-08
add product version to the download-file.json


vchrisb:
It would be good to add `product_version` to the `download-file.json` like `stemcell_version`.
e.g.
```
{
  "product_path": "p-healthwatch-1.5.2-build.4.pivotal",
  "product_slug": "p-healthwatch",
  "product_version": "1.5.2",
  "stemcell_path": "light-bosh-stemcell-170.51-google-kvm-ubuntu-xenial-go_agent.tgz",
  "stemcell_version": "170.51"
}
```

jtarchie:
Why is this information useful? What would you use it for?

vchrisb:
I would like to have first a download only pipeline which has a broader glob, like `2.5.*`, in the download-product config files. After downloading and storing the artifacts in s3, at the end of the pipeline all exact versions, like `2.5.3`, are committed to the download-product-s3 config files of the first stage, sandbox. The sandbox pcf deploy pipeline is triggered on that commit and does install/upgrade the known versions. Again at the end of the sandbox pipeline, the exact versions are committed to the download-product-s3 config files of the second stage, dev... and so on.
This allows to track in git the versions deployed and roll a specific set of product versions automatically through multiple stages.
Does that make sense?

kcboyle:
PR welcome, but we do have this prioritized. 

If you would like to start working on it, please tell us, and we'd love to work with you on this. 

To clarify, does this get used by `assign-stemcell` at all or is this strictly to optimize another workflow? 

We had discussed as a team maybe outputting an additional file (not called `assign-stemcell.json` that could have this value (and perhaps other values in the future that might be deemed useful). @vchrisb

vchrisb:
Happy to do a PR, as this should be easy. `DownloadProductOptions` already includes `ProductVersion` in https://github.com/pivotal-cf/om/blob/46ce73cb58cb29472d606b40de89ca771e317656/commands/download_product.go#L48
It would be used to as described in the previous comment and I think `download-file.json` is a good place.

kcboyle:
@vchrisb, sounds like that would be easy enough to pass through. 
My first thought was to grab the product version from the downloaded tile itself, but as that product can be opsman, a tile, a stemcell, or even a terraform template (or more!) that might get complicated really fast, and potentially be unreliable. 

Like I said though, if you have time and want to get it hashed out, feel free, but we do have this coming up in the next couple of weeks in our backlog.

are you thinking `download-file.json` as another outputted file?

[...]

Issue #362, opened 2019-05-10
[semi-random/use-case] hierarchical config and promotion versions


aegershman:
This isn't an issue; I was inspired by https://github.com/pivotal-cf/om/issues/360#issuecomment-490781915

I want to comment on @vchrisb 's use-case in [#360] without hijacking their issue. We do "hierarchical configuration monorepos"(?) && I want to share thoughts on how we promote config/versions from sandbox to dev to prod for all our IaaSes/regions/envs/foundations.

## current usage

Similar to what vchrisb described: we glob the `patch` version of pivnet product, download it to S3, then automation takes the _exact_ version of the latest that we just downloaded && opens a PR to a github repo, which starts this upgrade chain from env to env.

Here's a very stripped-down version of our `upgrade-tiles-config` configuration monorepository that holds information on the `product-version` and `stemcell-version` which should be applied for each foundation. I'm showing a limited set of configuration && limited subset of the `upgrade-tiles-config/` monorepo here:

```
pas
├── aws
│   └── us-west-2
│       ├── dev
│       │   ├── cf
│       │   │   ├── product-version
│       │   │   ├── stemcell-version
│       │   │   └── vars.yml
│       │   └── config
│       │       ├── auth.yml
│       │       ├── cf-secrets.yml
│       │       ├── director-secrets.yml
│       │       ├── env.yml
│       │       └── opsman.yml
│       ├── prod
│       │   ├── cf
│       │   │   ├── product-version
│       │   │   ├── stemcell-version
│       │   │   └── vars.yml
│       │   └── config
│       │       ├── auth.yml
│       │       ├── cf-secrets.yml
│       │       ├── director-secrets.yml
│       │       ├── env.yml
│       │       └── opsman.yml
│       └── sandbox
│           ├── cf
│           │   ├── product-version
│           │   ├── stemcell-version
│           │   └── vars.yml
│           └── config
│               ├── auth.yml
│               ├── cf-secrets.yml
│               ├── director-secrets.yml
│               ├── env.yml
│               └── opsman.yml
├── azure
│   └── centralus
│       ├── dev
│       │   ├── cf
│       │   │   ├── product-version
│       │   │   ├── stemcell-version
│       │   │   └── vars.yml
│       │   └── config
│       │       ├── auth.yml
│       │       ├── cf-secrets.yml
│       │       ├── director-secrets.yml
│       │       ├── env.yml
│       │       └── opsman.yml
│       └── sandbox
│           ├── cf
[...]

PR #368, opened 2019-05-21
add product_version in download-file.json

vchrisb:
as discussed in #360 this adds `product_version` to the `download-file.json`file.
I've tested it with downloading opsman, a tile and a stemcell.


kcboyle:
Thanks for the PR! We hope that including the product version will help folks in their workflows (thanks to @aegershman for a detailed related use case! #362

PR #378, opened 2019-06-05
Pivnet Token optional when using download source

### LINKED ISSUE #371: Make Pivnet Token optional when using download source s3
Flag `--pivnet-api-token` is currently mandatory even when using `--source s3`
https://github.com/pivotal-cf/om/blob/master/commands/download_product.go#L47

And when using `--source s3`, `om` does not validate if all required parameters for `s3` are present.
E.g. when `--s3-secret-access-key ` is missing following unspecific error is printed:
`could not execute "download-product": Key: 'S3Configuration.RegionName' Error:Field validation for 'RegionName' failed on the 'required' tag`

vchrisb:
Fixed #371 

* add validation to make `--pivnet-api-token` required if `source` is not specified and created test for it
* removed `--pivnet-api-token` from s3 download product acceptance test

### DIFF
--- a/acceptance/download_product_test.go
+++ b/acceptance/download_product_test.go
@@ -52,7 +52,6 @@ var _ = Describe("download-product command", func() {
 				tmpDir, err := ioutil.TempDir("", "")
 				Expect(err).ToNot(HaveOccurred())
 				command := exec.Command(pathToMain, "download-product",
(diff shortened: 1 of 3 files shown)


kcboyle:
Merged via command line. This should be included in the next release of `om`

PR #379, opened 2019-06-06
Removed redundant commands in readme

jhmcstanton:
Looks like the bot was borked and repeatedly appended the commands to the readme. 

No idea where the bot is configured, so not sure what happened. Looks like maybe someone turned it off since it stopped committing this afternoon.


jhmcstanton:
Readme broke the build D:

jpmcb:
There was a network error. No worries, it is not an automatic rejection

jpmcb:
Thanks for the catch, @jhmcstanton ! We found the culprit in our CI, and have fixed it so it doesn't happen again.

kcboyle:
This bot existed in om/ci/tasks/readme/task.yml

We changed the commit name to be more helpful

PR #386, opened 2019-06-16
Fix broken link to Platform Automation docs

evanfarrar:
Also, why is this caveat even here? Can we document all the commands in one place? Is Platform Automation is not updating this list of commands with commands they add? Are the commands that aren’t used by platform automation unvalidated WIP? Can we delete the list?

### DIFF
--- a/README.md
+++ b/README.md
@@ -20,7 +20,7 @@ You can find them at
 
 **NOTE**: Additional documentation for om commands
 used in Pivotal Platform Automation
-can be found in [Pivotal Documentation](docs.pivotal.io/platform-automation).
+can be found in [Pivotal Documentation](https://docs.pivotal.io/platform-automation).
 
 ## Versioning
 


jpmcb:
Great questions.

In our experience and research, we have found that there are operators out there who use `om` exclusively on a single workstation to interact with an Ops Man (similar to how you would use the bosh-cli). These operators may be using `om` on it's own without Platform Automation. We do not want to isolate those users. 

We have several stories in the back log to fix the `om` documentation as it has become stale in its current state. The readme in the root directory is based on `om -h` and is automatically updated in our pipeline. Therefore, it is the most up to date. But many of the other readme files have gone stale and are not automatically updated.

> Are the commands that aren’t used by platform automation unvalidated WIP?

No. If a command is a WIP or experimental, the help section of `om -h` will state: `**EXPERIMENTAL**`.

PR #387, opened 2019-06-17
Make UUID non configurable

tonyelmore:
AppDynamics tile was having the GUID being created as a configurable property and it should not have been.  This had previously been implemented in Tile-Config-Generator (https://github.com/pivotalservices/tile-config-generator/commit/40d4edb153a0caf67b7dafce9feea225e3ff0901)

Co-authored-by: Shawn Neal <sneal@pivotal.io>
Co-authored-by: Mike Hoskins <mhoskins@pivotal.io>


jtarchie:
Do they define the attribute in their metadata file to be non-configurable? That is what `optional` is for.

There might be some tiles that may want UUID to be configurable.

jtarchie:
@tonyelmore thanks for the PR.

Since a field is `configurable: false` be default.
We should fix `IsConfigurable` to just be `IsExplicityConfigurable` everywhere.
We've made that change and pushed it.

PR #389, opened 2019-06-19
bosh-env command does not resolve private key

jghiloni:
Right now, `om bosh-env -i keyname` does no checking if a private key
is valid or not. That, in and of itself, is not a big deal, but it also
doesn't create an absolute path to the key. This PR addresses that. When
specifying a key via `-i` it will attempt to open the file, and if it
is successful, will use its absolute path in `BOSH_ALL_PROXY` and
`CREDHUB_PROXY`. If the file is not openable, it uses just the provided
string, to provide backwards behavior compatibility.

### DIFF
--- a/commands/bosh_environment.go
+++ b/commands/bosh_environment.go
@@ -2,6 +2,8 @@ package commands
 
 import (
 	"fmt"
+	"os"
+	"path/filepath"
 	"regexp"
 	"strings"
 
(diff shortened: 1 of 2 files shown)


jtarchie:
Thanks for the PR. This brings an interesting discussion that we don't handle the key file not actually existing. If it does not exist, we should error out on that. Otherwise debugging your SSH session not working will be pretty hard.

We are going to discuss this.

jtarchie:
Thanks for the PR.
We added logic to error if the ssh key does not exist.

PR #393, opened 2019-06-20
Make sure Custom VM Types are created before resource config is speci…

jtarchie:
…fied,

so that custom VM Types can be used in director resource config.


jtarchie:
Thanks for the PR!

PR #394, opened 2019-06-24
Enforce minimum TLS version of 1.2

anyu:
- current Ops Manager implementations should support TLS 1.2


nbconklin:
thank you - we will look into this

jtarchie:
Thanks for the PR.

Does the CLI need to enforce this? If so, do any of the other CLIs in CF/Pivotal land share this logic.

OpsManager is enforcing this from the server side.

jtarchie:
It looks like the `cf-cli` is using min version of 1.0. https://github.com/cloudfoundry/cli/blob/b6a3695a5db06d6628428067e0a78e63041ea1ca/cf/net/ssl.go#L10

We need better guidance on _standards_ here.

PR #397, opened 2019-06-29
Run go mod tidy.

evanfarrar:
I was using `om` as a library and `kiln` also as a library and got really confused when my version of kiln was a spring-2018 vintage. It's weird to have a dependency management system that tracks things that aren't dependencies ¯\_(ツ)_/¯ but we love you anyway for some reason golang.


jtarchie:
Thanks for point it out.
A commit was made with your name. :-)

PR #405, opened 2019-07-22
Added `revert-staged-changes` to index README

eamonryan:
`revert-staged-changes` is present in newer versions of `om` but was just missing from the main README.

### DIFF
--- a/README.md
+++ b/README.md
@@ -163,6 +163,7 @@ Commands:
   pending-changes                 lists pending changes
   pre-deploy-check                **EXPERIMENTAL** lists pending changes
   regenerate-certificates         deletes all non-configurable certificates in Ops Manager so they will automatically be regenerated on the next apply-changes
+  revert-staged-changes           reverts staged changes on the Ops Manager targeted
   ssl-certificate                 gets certificate applied to Ops Manager
   stage-product                   stages a given product in the Ops Manager targeted
   staged-config                   **EXPERIMENTAL** generates a config from a staged product


jtarchie:
It has actually been [officially removed](https://github.com/pivotal-cf/om/commit/ee9a0885ce937c31474d765a9991270ff96c8616) from `om` as it has been reported it did not work with newer versions of OpsManager.

Where did you see it in _newer versions_?

eamonryan:
@jtarchie It looks like it was only just removed in om 3.0.0, but there's almost no notes at https://github.com/pivotal-cf/om/releases/tag/3.0.0 so I didn't see it documented.

Additionally, the commit you linked says that `unstage-product` does the same thing - what if you want to revert the BOSH tile's config? (e.g. after manually filling it out, running `staged-director-config` and wanting to clear the config to be sure you have all properties) - does `unstage-product` work for that too?

jtarchie:
It may not have the full feature set as `revert-staged-changes`. But this command did not work, anyways. It was relying on submitting an HTML form from the Web UI. There is not API endpoint that we know of _reverting changes_.

What use case do you have for reverting changes via automation? Aren't you able to apply the older configuration file?

eamonryan:
@jtarchie I wasn't doing the revert via pipeline, but was writing out some steps which included:
- Manually configure BOSH tile
- Run `staged-director-config`
- Revert manual changes to ensure config is clear
- Run `configure-product`

So was just including `om revert-staged-changes` as part of that, to help ensure no properties were missed, but I can leave it as a "You must do this in the UI" step!

nbconklin:
@eamonryan we have an `unstage-product` in `om` that allows you to do this from the API. Closing this for now, feel free to re-open if needed!

PR #406, opened 2019-07-23
updated reference command to correct flag

sapient007:
OM CLI uses --username instead of --user for configure-authentication command. this update corrects the documentation to show proper flag


kcboyle:
seems like an oversight. Thanks for the PR!

PR #410, opened 2019-08-13
Add --force-vm-extensions to configure-product

### LINKED ISSUE #400: Setting `additional_vm_extensions` to `[]` should erase all extensions.
There's currently no way to remove all VM Extensions from a VM using the `resource-config` block of a config yaml. I would expect setting `additional_vm_extensions: []` would have that effect, but it's the same effect as not setting the key at all.

jghiloni:
If set, this flag will delete all VM extensions from a given job's resource config, if additional_vm_extensions` is explicitly set to `[]`. If the flag or `additional_vm_extensions` is missing, then the existing VM Extensions will not be affected.

Resolves #400.


anEXPer:
Went back to issue #400 to describe what we've landed on for addressing that issue; we'd rather have the config be the source of truth more or less as originally described, rather than having a flag (which, we know, we suggested) complicate the issue.

I'm closing this PR, though it looks like fine work. We'd be happy to take a similar PR without the flag, matching the behavior I described in issue #400.

PR #412, opened 2019-08-14
Rename `tile-metadata` to `product-metadata`

### LINKED ISSUE #275: `tile-metadata` should probably be `product-metadata`
The `tile-metadata` command is the only place in om we refer to products as tiles. It even takes a `--product-path`!

We should rename it, but this would be a breaking change.

jghiloni:
This PR resolves #275, and effectively aliases `tile-metadata` to `product-metadata` and
adds a deprecation warning to the usage of `tile-metadata`.


kcboyle:
Thanks for the PR @jghiloni !

PR #433, opened 2019-10-02
Allow alternatives for config-template cli args

### LINKED ISSUE #432: Allow config-template to take a config file in lieu of command line flags
Because `config-template` requires sensitive data (the user's Pivnet token) to be specified on the command line, it should be able to accept a config file just like `download-product` or like `--env` to bypass that. I discussed this with @kcboyle today and she said she'd be willing to accept a PR, which I'll submit referencing this issue. A full set of tests will accompany the PR.

jghiloni:
This resolves #432. Because `download-product` can accept required
arguments in various ways (on the command line, via vars file, vars,
or env variables) and because `config-template` has as its options
a subset of `download-product`'s argumennts, it makes sense to allow
`config-template` to accept arguments in the same fashion. This PR
refactors `config-template` to use the same common code as
`download-product` and includes the same tests as well. In addition,
`commands.NewConfigTemplate` needs to accept any `environFunc` instead
of hardcoded `os.Environ`, so a second `New*` method was introduced to
ensure backward compatibility.

### DIFF
--- a/commands/config_template.go
+++ b/commands/config_template.go
@@ -11,9 +11,14 @@ import (
 )
 
 type ConfigTemplate struct {
-	environFunc   func() []string
+	environFunc   envProvider
 	buildProvider buildProvider
 	Options       struct {
+		ConfigFile string   `long:"config"                     short:"c" description:"path to yml file for configuration (keys must match the following command line flags)"`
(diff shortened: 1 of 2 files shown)


kcboyle:
thanks for the PR!

PR #444, opened 2019-10-28
Return stemcell_library information

yuyangbj:
The API stemcell_assiciation will return stemcell_library information,
change the interface to support it.

### DIFF
--- a/api/multi_stemcell_service.go
+++ b/api/multi_stemcell_service.go
@@ -9,6 +9,7 @@ import (
 
 type ProductMultiStemcells struct {
 	Products []ProductMultiStemcell `json:"products"`
+        StemcellLibrary []StemcellObject `json:"stemcell_library"`
 }
 
 type StemcellObject struct {


jtarchie:
Hi @yuyangbj.

Can you tell us more what this PR is for?

yuyangbj:
@jtarchie thanks for the quick response, I want to query some specific stemcells in my automation process. For example in Pivotal Container Service, i want to automate the windows plan creation, i need to query if windows stemcell is existing or not before I configure the PKS product with windows plan enabled.

jtarchie:
I'm not sure I understand.
Will this affect one of our commands in `om` that provides more output?

yuyangbj:
@jtarchie i do not want to get more output for om cmd, i just want to leverage the rest api. But actually the om is missing a CLI to list stemcells, do you think so?

jtarchie:
We make no guarantees about the code related to the `api` layer. The `om` code is meant to interfaced through the commands. If the commands don't require it, we won't add extraneous fields to response payload. I'm going to close this PR based on that.

As for the command to list stemcells, there hasn't been a use case to add it yet. Please open up an issue to discuss it if you'd like to add one and why it would be helpful to list the stemcells.

yuyangbj:
@jtarchie i opened an issue https://github.com/pivotal-cf/om/issues/445, we can discuss there

PR #464, opened 2020-02-08
Add tests for om update-ssl-certificate with config/vars

### LINKED ISSUE #463: update-ssl-certificate should support files and quoted strings
It is unclear that you need to provide `--certificate-pem` and `--private-key-pem` as command line objects that are unquoted to `update-ssl-certificate`

It would be a nice user experience if these could be passed in as files and/or passed in as quoted strings.

jghiloni:
This closes #463. The code to implement this was included in commit 1995e93cda087451d3330f31756b1e21283931ea but tests were not, so this should cover them.

### DIFF
--- a/commands/loadConfigFile.go
+++ b/commands/loadConfigFile.go
@@ -2,11 +2,12 @@ package commands
 
 import (
 	"fmt"
+	"reflect"
+	"strconv"
+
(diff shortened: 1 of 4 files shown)


kcboyle:
merged via command line. Thanks for the PR @jghiloni !

Issue #488, opened 2020-04-24
Feature Request: bosh-diff should exit non-zero if changes exist


jghiloni:
The standard GNU `diff` tool exits with 1 if there are differences, and 0 if the files are identical. I'd like to propose doing the same with `om bosh-diff`, as that could make things much easier to script. An example would be if there are no changes to a deployment, you could skip the `apply-changes` altogether. I'd be happy to contribute a PR if this is something that would be of interest.

kcboyle:
makes sense to me

I'd make a suggestion to make the exit 1 an option (like in `pending-changes`, it uses the `--check` flag). 

Happy to accept a PR.

jghiloni:
I like that! I'll put together the PR today

kcboyle:
Merged PR! closing the issue!

kcboyle:
Looks like there might be something missing. When making changes to the bosh director (not a product) it looks like `--check` doesn't return a proper exit status 
```
$ ./om -t https://pcf.sapphireblue.cf-app.com -u pivotalcf -p jt2blnjg6n4xz3ek bosh-diff --check
## Director Manifest
 instance_groups:
 - name: bosh
   properties:
     director:
+      workers: "<redacted>"
-      workers: "<redacted>"
+      max_threads: "<redacted>"
-      max_threads: "<redacted>"
## Director Cloud Config
no changes
## Director Runtime Configs
no changes
## Director CPI Configs
no changes
## Product Manifest for cf
no changes
## Runtime Configs for cf
no changes
$echo $?
0
```

versus with a product change
```
±  |master ?:1 ✗| → ./om -t https://pcf.sapphireblue.cf-app.com -u pivotalcf -p jt2blnjg6n4xz3ek bosh-diff --check
## Director Manifest
 instance_groups:
 - name: bosh
   properties:
     director:
+      workers: "<redacted>"
-      workers: "<redacted>"
+      max_threads: "<redacted>"
-      max_threads: "<redacted>"
## Director Cloud Config
no changes
## Director Runtime Configs
no changes
## Director CPI Configs
no changes
## Product Manifest for cf
 instance_groups:
 - name: mysql_monitor
   jobs:
   - name: replication-canary
     properties:
       mysql-monitoring:
         replication-canary:
+          poll_frequency: 90
-          poll_frequency: 30
## Runtime Configs for cf
no changes
could not execute "bosh-diff": Differences exist between the staged and deployed versions of the requested products
$echo $?
1
```

[...]

PR #490, opened 2020-04-30
Add check flag to bosh-diff so it can be used in automation

jghiloni:
This implements the feature requested in #488 

Allow users to specify the `--check` command when using `om bosh-diff` to get a non-zero exit code when changes exist, like linux `diff` and `om pending-changes`. This will allow automation users to take action depending on whether or not changes exist.

### DIFF
--- a/commands/bosh_diff.go
+++ b/commands/bosh_diff.go
@@ -2,10 +2,11 @@ package commands
 
 import (
 	"fmt"
-	"github.com/fatih/color"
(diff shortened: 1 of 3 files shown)


kcboyle:
thanks for the PR!

kcboyle:
Some comments:
make sure to test (a piece of) the error message
remember to run docsgenerator when adding new flags
remember to update the changelog with any new features!

PR #501, opened 2020-07-14
Do not require the id arg for certificate-authority if there is only one CA to choose from

jghiloni:
It feels extraneous to have to run `om certificate-authorities` to get the CA's ID if there's only one. This changes that behavior, but preserves existing behavior in all other cases


kcboyle:
Merged! will be released in `om` 6.0

Issue #207, opened 2018-07-19
om configure-product should be enhanced to support collections 


sadvani:
Configuring or Editing collections in OpsManager via the API behaves in a non deterministic manner. Some service tiles (MySQL v2, maybe others) use collections to define Service Plans in the tile. Currently, if a collection has multiple records, and a change is being made to one collection record by passing on only the GUID of the record in question along with the change, OpsManager will delete all other collection records. This means that only the record that the Operator was trying to edit will be saved. Additionally, if an operator tries to change a collection record without specifying a GUID, OpsManager will proceed to delete all records and create a new one with a new GUID. This issue has been seen a few times with the MySQL Tile [1]. We flagged this issue with the Pcf-Pipelines and PAS teams and they suggested we open an issue in this repo [2]

A workaround for this issue is for om configure-product to send all GUIDS and plans on every PUT. 

[1] https://github.com/pivotal-cf/pcf-pipelines/issues/332
[2] https://pivotal.slack.com/archives/C0554UNSS/p1531941382000046

sadvani:
I believe it does always fail. You could either use the MySQL v2 tile or the OpsManager example product. In the MySQL case, you will have to deploy the entire once and then use configure-product and deploy again. This deploy should fail because the GUIDs associated with the deployed service plans will have changed. 

You can use the example tile by just using configure-product and noticing the strange behavior of records getting deleted, GUIDs changing, etc.

mcwumbly:
Cool, I'll assume that's the case. We'll try to prioritize this ourselves at some point if no one else gets to it first, but I'm throwing a `pr-welcome` on it in the meantime.

jtarchie:
@sadvani and @mcwumbly, are you able to review the PR from @calebwashburn. I've been able to see that the code works, but I don't have a way (or knowledge) on how reproduce this issue.

sadvani:
@mcwumbly @jtarchie @calebwashburn is this issue fixed?

calebwashburn:
@sadvani Believe so

fredwangwang:
yes, this pr https://github.com/pivotal-cf/om/pull/227 has been pulled in. We validated using @calebwashburn's gist. Do you still see a problem?

sadvani:
Perfect, thanks so much @calebwashburn @fredwangwang. 

Is there a specific version of OM that I should tell the customer to upgrade to?

fredwangwang:
master, there is no release cut since the fix introduced.

[...]

PR #502, opened 2020-08-04
Use name, key, *name$ logical keys to find existing property GUIDs

mrdavidlaing:
Extends `configure-product`'s _decorating collection with guid based on `name` logical key_ logic to also use:

* `key` or 
* fields ending in `name` (eg, `sqlServerName`)

as logical keys.  

Partially addresses [#207](https://github.com/pivotal-cf/om/issues/207)


api/staged_products_property_collection_helpers.go
```diff
@@ -0,0 +1,142 @@
...
+		return value
+	}
+	return ""
+}
+
+func (item updatedPropertyCollectionItem) setFieldValue(fieldName string, value string) {
+	item.Data[fieldName] = value
+}
+
+func parseUpdatedPropertyCollection(updatedProperty interface{}) (updatedPropertyCollection, error) {
```
kcboyle, line 25: In general, we try to avoid nested if statements in go. This helps with readability and maintainability as well, though it takes more lines of code.

Alternatively, this function could be written like
```
func parseUpdatedPropertyCollection(updatedProperty interface{}) (updatedPropertyCollection, error) {
	var collection updatedPropertyCollection

	updatedPropertyAsMap, ok := updatedProperty.(map[string]interface{})
	if !ok {
		return nil, fmt.Errorf("parseUpdatedPropertyCollection: failed to convert %v to map[string]interface{}", updatedProperty)
	}

	rawItems := updatedPropertyAsMap["value"]

	rawItemSlice, ok := rawItems.([]interface{})
	if !ok {
		return nil, fmt.Errorf("parseUpdatedPropertyCollection: failed to convert %v to []interface{}", rawItems)
	}

	for _, item := range rawItemSlice {
		itemMap, ok := item.(map[string]interface{})
		if !ok {
			return nil, fmt.Errorf("parseUpdatedPropertyCollection: failed to convert %v to map[string]interface{}", item)
		}
		
		collection = append(collection, updatedPropertyCollectionItem{Data: itemMap})
	}
	
	return collection, nil
}
```

api/staged_products_property_collection_helpers.go
```diff
@@ -0,0 +1,142 @@
...
+func (item responsePropertyCollectionItem) getFieldValue(fieldName string) string {
+	if valueObj, ok := item.Data[fieldName].(map[interface{}]interface{}); ok {
+		if value, ok := valueObj["value"].(string); ok {
+			return value
+		}
+	}
+	return ""
+}
+
+//Find and return first string matching regex pattern
```
kcboyle, line 77: Rather than a comment, I think I'd prefer to name this function something like `findFirstMatch`

api/staged_products_property_collection_helpers.go
```diff
@@ -0,0 +1,142 @@
...
+//returns matched <string, true> if found, or <"",false> if not
+func findFirst(strings []string, fieldRegex string) (string, bool) {
+	r, err := regexp.Compile(fieldRegex)
+	if err != nil {
+		panic(fmt.Sprintf("failed to compile regex: %v because:\n%v", fieldRegex, err))
```
kcboyle, line 82: we shouldn't panic in the code (even if the code shouldn't fail). We're going to have this function return an error, and exit based on the error instead

api/staged_products_property_collection_helpers.go
```diff
@@ -0,0 +1,142 @@
...
+			return str, true
+		}
+	}
+	return "", false
+}
+
+//Finds logical key field (if exists)
+//returns <fieldName, true> if map contains field that can be considered a logical key
+//returns <"", false> if no logical key found
+func (item responsePropertyCollectionItem) findLogicalKeyField() (string, bool) {
```
kcboyle, line 95: how this function is used in associateExistingCollectionGUIDs doesn't require the bool return type. Since there's no opportunity for error, it can just return the found string.

api/staged_products_property_collection_helpers.go
```diff
@@ -0,0 +1,142 @@
...
+	//Extract & sort the fieldnames to ensure search is deterministic (since map order isn't guarenteed)
+	sortedFields := make([]string, 0, len(item.Data))
+	for k := range item.Data {
+		sortedFields = append(sortedFields, fmt.Sprintf("%v", k))
+	}
+	sort.Strings(sortedFields)
+
+	if fieldName, ok := findFirst(sortedFields, "^name$"); ok { //First look for a field named 'name'
```
kcboyle, line 103: instead of a list of if/else statements that call a function used only in this method, we can create a slice (essentially an array) of the regexes, and then loop over those and call that function in the for loop. for example:
```
regexes := []string{"^name$", "^key$", "(?i)name$"}
	for _, regex := range regexes {
		compiledRegex := regexp.MustCompile(regex)

		for _, str := range sortedFields {
			if compiledRegex.MatchString(str) {
				return str
			}
		}
	}
```

api/staged_products_property_collection_helpers.go
```diff
@@ -0,0 +1,142 @@
+package api
```
kcboyle, line 1: So I've been tossing back and forth between how I feel about the 'helpers' method. Since this is just for collections, I think I'd prefer this to be named `collections_service.go`.

api/staged_products_property_collection_helpers_tests.go
```diff
@@ -0,0 +1,283 @@
+package api //not in api_tests because we are intentionally testing some functionality internal to the package
```
kcboyle, line 1: Same as the other comment, but `collections_service_test.go` (not plural. if you're using a clever IDE like Goland, it won't recognize the auto-open test if a plural as well)

api/staged_products_property_collection_helpers.go
```diff
@@ -0,0 +1,142 @@
...
+	}
+	for _, str := range strings {
+		if r.MatchString(str) {
+			return str, true
+		}
+	}
+	return "", false
+}
+
+//Finds logical key field (if exists)
```
kcboyle, line 92: This comment is unnecessary. It was helpful for the review, but you have this tested, so nix the comments!

api/staged_products_property_collection_helpers.go
```diff
@@ -0,0 +1,142 @@
...
+	} else if fieldName, ok := findFirst(sortedFields, "^key$"); ok { //then a field named 'key'
+		return fieldName, true
+	} else if fieldName, ok := findFirst(sortedFields, "(?i)name$"); ok { // otherwise a field that ends with 'name' (case insensitive)
+		return fieldName, true
+	}
+
+	return "", false
+}
+
+func (collection responsePropertyCollection) findLogicalKeyFieldFromFirstCollectionItem() (string, bool) {
```
kcboyle, line 114: while I appreciate the readability of this helper function, it is easy to read as `logicalKeyFieldName := existingCollection[0].findLogicalKeyField()` without the helper function in the below method

api/staged_products_property_collection_helpers.go
```diff
@@ -0,0 +1,142 @@
...
+	return collection[0].findLogicalKeyField()
+}
+
+//Find and associate the GUID for those collection items that already exist in OpsMgr
```
kcboyle, line 118: nix the comment. the function name is fine

api/staged_products_property_collection_helpers.go
```diff
@@ -0,0 +1,142 @@
...
+	if err != nil {
+		return err
+	}
+	existingCollection, err := parseResponsePropertyCollection(existingProperty.Value)
+	if err != nil {
+		return err
+	}
+	if logicalKeyFieldName, ok := existingCollection.findLogicalKeyFieldFromFirstCollectionItem(); ok {
+		//Use the logical key to find associated GUIDs (from the existingProperty collection items) and assign them to the updatedProperty item
+		for _, updatedCollectionItem := range updatedCollection {
```
kcboyle, line 131: we thought maybe this could be wrapped in a function to better describe what it's doing; something like:
```
func assignExistingGUIDtoPropertyItem(existingCollection responsePropertyCollection, updatedCollection updatedPropertyCollection, logicalKeyFieldName string) {
	for _, updatedCollectionItem := range updatedCollection {
		updatedCollectionItemLogicalKeyValue := updatedCollectionItem.getFieldValue(logicalKeyFieldName)
		for _, existingCollectionItem := range existingCollection {
			if updatedCollectionItemLogicalKeyValue == existingCollectionItem.getFieldValue(logicalKeyFieldName) {
				updatedCollectionItem.setFieldValue("guid", existingCollectionItem.getFieldValue("guid"))
				break //move onto the next updatedCollectionItem
			}
		}
	}
}
```

kcboyle:
Committed some of the proposed changes to a [branch](https://github.com/pivotal-cf/om/tree/pr-502-updates) if you'd like to give it a look over.

api/staged_products_property_collection_helpers.go
mrdavidlaing: Done.

Makes for an interesting inversion of the `try { happyPath() } catch exception {}` flow I'm used to when reading the code; but removing the nesting is a big plus

api/staged_products_property_collection_helpers_tests.go
mrdavidlaing: Good to know.  Thanks!

api/staged_products_property_collection_helpers.go
mrdavidlaing: I felt dirty doing this; but went ahead since I felt this was an error that could only occur at "compile time" (by the programmer using the function passing in "bad" regex)

The refactoring of `findLogicalKeyField()` proposed below sidesteps this issue; since the regexes in question are in the same function.  (Interestingly; I believe that `MustCompile()` will also `panic` if the regex is invalid - https://golang.org/src/regexp/regexp.go?s=10768:10804#L298)

api/staged_products_property_collection_helpers.go
mrdavidlaing: That is much better - thanks!  Implemented in 34f7985

The order in which we sort is important - [test](https://github.com/pivotal-cf/om/pull/502/commits/9e73e245a6467a586e74af526c8e3a94377db414#diff-cb697bc504ae3e69003a9c246d254be0R250) - so I left in [a comment calling this out](https://github.com/pivotal-cf/om/pull/502/commits/34f7985866373f2ddbe5b34ca3e32bb16ffc0e1d#diff-05a0e719066db6ea2a4255f414b423ebR100).

api/staged_products_property_collection_helpers.go
mrdavidlaing: Agreed.

However, refactoring of findLogicalKeyField() - 34f7985 - sidesteps this issue.

api/staged_products_property_collection_helpers.go
mrdavidlaing: Clean Code FTW. 

Especially since the next PR I'm intending on submitting with be the implementation of [`assignExistingGUIDUsingValueHash()`](https://docs.google.com/document/d/1ZWX7twODLiyGTXnYP6xWtJZnrkLaJw2VTRBAUkbdz5I/edit#heading=h.z3pgxfm4iuvk)

Implemented in 9541300

api/staged_products_property_collection_helpers.go
mrdavidlaing: I wanted to call out the assumption that the all the collection items will have the same logical key - something I believe is true; but isn't explicitly enforced anywhere.

Removed the function as suggested (in 9541300); but left in a comment calling out the assumption.

api/staged_products_property_collection_helpers.go
mrdavidlaing: Whilst true currently; *not* finding a logical key (because there isn't one) is information we will need in the future when implementing _assign GUIDs using some other strategy_ logic.

Since checking a bool return type seems cleaner than checking for an empty string; I'd prefer to leave this as is for now.

I've attempted to capture this requirement in [this test](https://github.com/pivotal-cf/om/pull/502/commits/9e73e245a6467a586e74af526c8e3a94377db414#diff-cb697bc504ae3e69003a9c246d254be0R160)

api/staged_products_property_collection_helpers.go
mrdavidlaing: Done in 4bab125

api/staged_products_property_collection_helpers.go
mrdavidlaing: Done in 4bab125

api/staged_products_property_collection_helpers.go
mrdavidlaing: Happy to get rid of smelly `helpers` name.

However; I'm not sure `collections_service` is better; for two reasons:

1.  It implies a more `generic` set of functionality than is implemented - which is very specific to the (complicated) way the `staged_products` API works.  I think we should keep a name that includes this specificity (eg: `staged_products_property_collection_<something>`
2.  It's quite different to all the other `_service` files in that it doesn't make any network calls.

What about `staged_products_property_collection_GUID_association` or `staged_products_property_collection_GUID_assigner`?

kcboyle: fair points.
optimally, I'd love to see all of those things just in the `staged_config_service`, but recognize that this might be a bit of work and a little harder to test the more intricate functionality. These functions are _only_ used in `staged_config`, so even if we rename the file, it still feels a little weird. 

given the choice, I really like `staged_products_property_collection_guid_assigner` name (lowercase instead of uppercase guid in the name)

mrdavidlaing: Renamed to `staged_products_property_collection_guid_assigner` by e9c69d6

api/staged_products_property_collection_helpers_tests.go
mrdavidlaing: Addressed in e9c69d6

api/staged_products_property_collection_helpers.go
mrdavidlaing: Renamed to `staged_products_property_collection_guid_assigner` by e9c69d6

kcboyle:
Looks good to @jtarchie and I. Thanks @mrdavidlaing !

PR #503, opened 2020-08-07
Use equivalent item values to find existing property GUIDs

mrdavidlaing:
Related to #502, this addition extends the _decorating collection with guid_ logic to associate collection item guids with items that have equivalent item.

For example (see [test](https://github.com/pivotal-cf/om/blob/8a43e6df7a8ce99414c59e2513e951d73832f691/api/staged_products_service_test.go#L644)); if the following item exists in an OpsMgr Property Collection:

```
  { 
    "guid": "28bab1d3-4a4b-48d5-8dac-two",
    "another_property": true,
    "a_property": "\"value\" of 'second' item"
  }
```

And an operator uses `om configure-product` to "update" that property with no changes (as might be the case in a GitOps workflow):

```
  { 
    "a_property": "\"value\" of 'second' item"
    "another_property": true,
  }
```

`om configure-product` will decorate the updated property with the `guid` of the existing equivalent item (ignoring key order); which would register (correctly) as a `no-op` change by the `om bosh-diff` functionality; preserving the existing item `guid`

This partially addresses #207

_Known issues_
* Does not associate GUIDs for collections that have a secret field AND no logical key

### DIFF
--- a/api/staged_products_property_collection_guid_assigner_test.go.go
+++ b/api/staged_products_property_collection_guid_assigner_test.go.go
@@ -1,6 +1,7 @@
 package api //not in api_tests because we are intentionally testing some functionality internal to the package
 
 import (
+	"encoding/json"
 	"fmt"
 
 	. "github.com/onsi/ginkgo"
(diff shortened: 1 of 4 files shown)


api/staged_products_property_collection_guid_assigner_test.go.go
```diff
@@ -9,17 +10,25 @@ import (
 )
 
 var _ = Describe("ResponsePropertyCollection", func() {
-	unmarshalJSON := func(json string) interface{} {
+	unmarshalJSONLikeApiGetStagedProductProperties := func(json string) interface{} {
```
mrdavidlaing, line 13: Out of interest; why does [`Api.GetStagedProductProperties`](https://github.com/pivotal-cf/om/blob/d17b3b7de0e3c435ea14951bb05dbc0616b95a93/api/staged_products_service.go#L323) use `yaml.Unmarshal()` rather than `json.Unmarshal()`?

kcboyle: No particular reason. We inherited this codebase from the releng team, and haven't made everything consistent yet

kcboyle:
This looks good to us. Stylistically we don't see any problems. We're going to merge in and test. Thanks for the PR!

api/staged_products_property_collection_guid_assigner_test.go.go
mrdavidlaing: Good to know.  

My 2c - we should switch to `json.Unmarshal`; since it produces more usable `map[string]interface{}` interfaces.

PR #504, opened 2020-08-27
BUGFIX: Ignores non-configurable properties when associating existing collection item GUIDs 

mrdavidlaing:
Fixes a bug (related to #207) where new GUIDs were being assigned for collection items that contained non-configurable properties even though non of the values had changed


api/staged_products_property_collection_guid_assigner.go
```diff
@@ -115,6 +115,9 @@ func (item responsePropertyCollectionItem) getFieldValuesExceptGUID() map[interf
 		if key == "guid" {
 			continue
 		}
+		if !valueObj.(map[interface{}]interface{})["configurable"].(bool) {
```
kcboyle, line 118: In terms of readability, the only think I would change is pulling out `!valueObj.(map[interface{}]interface{})["configurable"].(bool)` into
```
isNotConfigurable = valueObj.(map[interface{}]interface{})["configurable"].(bool)
if isNotConfigurable {
  continue
}
```

kcboyle:
I think this looks great. I'd be happy to pull it in. CI failures are due to things not in your control

api/staged_products_property_collection_guid_assigner.go
mrdavidlaing: Implemented.

PR #505, opened 2020-08-27
Handle collections containing secrets when finding existing property GUIDs 

mrdavidlaing:
Addresses one of the known issues for collection GUID assignment addressed in #207:
 
* [GUID association does not work for collections items that have a secret field AND no logical key field](https://github.com/pivotal-cf/om/issues/207#issuecomment-672827665)

### DIFF
--- a/api/staged_products_property_collection_guid_assigner.go
+++ b/api/staged_products_property_collection_guid_assigner.go
@@ -115,6 +115,10 @@ func (item responsePropertyCollectionItem) getFieldValuesExceptGUID() map[interf
 		if key == "guid" {
 			continue
 		}
+		isNotConfigurable := !valueObj.(map[interface{}]interface{})["configurable"].(bool)
+		if isNotConfigurable {
+			continue
+		}
 		extractedValues[key] = valueObj.(map[interface{}]interface{})["value"]
 	}
 
--- a/api/staged_products_property_collection_guid_assigner.go
+++ b/api/staged_products_property_collection_guid_assigner.go
@@ -79,18 +79,37 @@ type responsePropertyCollectionItem struct {
 	Data map[interface{}]interface{}
 }
 
-func parseResponsePropertyCollection(rawItems interface{}) (responsePropertyCollection, error) {
+func parseResponsePropertyCollection(input associateExistingCollectionGUIDsInput) (responsePropertyCollection, error) {
 	var collection responsePropertyCollection
 
-	rawItemSlice, ok := rawItems.([]interface{})
+	rawItemSlice, ok := input.ExistingProperty.Value.([]interface{})
 	if !ok {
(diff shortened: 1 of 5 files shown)


api/staged_products_property_collection_guid_assigner.go
```diff
@@ -137,12 +156,20 @@ func (existingCollection responsePropertyCollection) findGUIDForEquivalentlItem(
 	return "", false
 }
 
-func associateExistingCollectionGUIDs(updatedProperty interface{}, existingProperty ResponseProperty) error {
-	updatedCollection, err := parseUpdatedPropertyCollection(updatedProperty)
+type associateExistingCollectionGUIDsInput struct {
```
mrdavidlaing, line 159: Most of the changes are a result of this refactoring, implemented to make passing the properties needed to enable `func parseResponsePropertyCollection()` to make a `Api.GetDeployedProductCredential()` call to fetch the existing secret values

api/staged_products_property_collection_guid_assigner.go
```diff
@@ -79,20 +79,39 @@ type responsePropertyCollectionItem struct {
...
 	}
 
-	for _, item := range rawItemSlice {
+	for index, item := range rawItemSlice {
 		itemMap, ok := item.(map[interface{}]interface{})
 		if !ok {
 			return nil, fmt.Errorf("parseResponsePropertyCollection: failed to convert %v to map[interface{}]interface{}", item)
 		}
 
+		for collectionItemKey, collectionItemObj := range itemMap {
```
mrdavidlaing, line 96: All the new logic is here

api/staged_products_property_collection_guid_assigner.go
```diff
@@ -79,20 +79,39 @@ type responsePropertyCollectionItem struct {
...
+			apiOutput, err := input.APIService.GetDeployedProductCredential(GetDeployedProductCredentialInput{
+				DeployedGUID:        input.ProductGUID,
+				CredentialReference: credentialName,
+			})
+			if err != nil {
+				return nil, err
+			}
+
+			collectionItemObjAsMap["value"] = apiOutput.Credential.Value
+		}
```
mrdavidlaing, line 113: All the new logic is here

mrdavidlaing: Fetching the credential values like this increases the HTTP requests made by  `om configure-product  --config example-product.config.yml`  from 40 to 60

kcboyle: is it going to be 60 requests every time, or just for collection objects?

kcboyle: not too worried about the time addition

kcboyle:
It looks like there's a merge conflict in `api/staged_products_property_collection_guid_assigner_test.go.go` (and it also has the `.go` extension twice, which seems weird). Other than that, LGTM

PR #511, opened 2020-09-02
Make `om bosh-diff` return with exit code 2

jghiloni:
Return 2 if `--check` is specified and there are differences instead
of 1 for all other errors.

Co-Authored-By: Amber Salome <asalome@vmware.com>


jghiloni:
Solved by #512

PR #512, opened 2020-09-02
Exit 2 on bosh-diff when changes exist and --check is set

jghiloni:
This obviates #511 

Co-Authored-By: Amber Salome <asalome@vmware.com>


jghiloni:
never mind

PR #519, opened 2020-10-26
Add --bosh and --credhub options for bosh-env command

### LINKED ISSUE #458: om bosh-env should be able to independently set credhub and bosh vars
See #385 for more context

A nice to have feature for `bosh-env` would be to allow bosh and credhub vars to be set separately. A use case for this would be if a user needed to interact with bosh using a credhub that was not deployed via bosh (or vice versa). Right now, this would require the user to manually unset the credhub vars and/or reset the existing credhub vars that they desired. 

We shouldn't break the existing behavior of `bosh-env`, so when running `om bosh-env`, it should, by default, include credhub and bosh env vars. 

The following is a first pass as to how this could look. It is open to discussion
To exclude credhub vars
```bash
om bosh-env --exclude-credhub
```

To exclude bosh vars
```bash
om bosh-env --exclude-bosh
```

I would expect that these could be used together, but nothing would be set if both flags were passed (perhaps with a message saying "nothing to do"), but it is also understandable that passing both flags would produce an error.

Thank you for the proposal @aegershman!

iplay88keys:
- Allows for printing of only the BOSH or Credhub environment variables

Resolves #458

### DIFF
--- a/commands/bosh_environment.go
+++ b/commands/bosh_environment.go
@@ -21,6 +21,8 @@ type BoshEnvironment struct {
 	opsmanHost      string
 	Options         struct {
 		ShellType     string `long:"shell-type" description:"Prints for the given shell (posix|powershell)"`
+		BoshVars      bool   `long:"bosh" short:"b" default:"false" description:"Prints the BOSH director environment variables"`
+		CredhubVars   bool   `long:"credhub" short:"c" default:"false" description:"Prints the Credhub environment variables"`
 		SSHPrivateKey string `long:"ssh-private-key" short:"i" description:"Location of ssh private key to use to tunnel through the Ops Manager VM. Only necessary if bosh director is not reachable without a tunnel."`
 	}
 }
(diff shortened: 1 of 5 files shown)


kcboyle:
Some output testing -- output order does not matter. This is existing behavior for `om bosh-env`
```
→ om --env env.yml bosh-env
export CREDHUB_SECRET=secret
export CREDHUB_SERVER=https://10.0.0.10:8844
export CREDHUB_CA_CERT='-----BEGIN CERTIFICATE-----
-----END CERTIFICATE-----
'
export BOSH_CLIENT=client
export BOSH_CLIENT_SECRET=secret
export BOSH_ENVIRONMENT=10.0.0.10
export BOSH_CA_CERT='-----BEGIN CERTIFICATE-----
-----END CERTIFICATE-----
'
export CREDHUB_CLIENT=client

→ om --env env.yml bosh-env --credhub
export CREDHUB_CLIENT=client
export CREDHUB_SECRET=secret
export CREDHUB_SERVER=https://10.0.0.10:8844
export CREDHUB_CA_CERT='-----BEGIN CERTIFICATE-----
-----END CERTIFICATE-----
'

→ om --env env.yml bosh-env --bosh
export BOSH_CLIENT=client
export BOSH_CLIENT_SECRET=secret
export BOSH_ENVIRONMENT=10.0.0.10
export BOSH_CA_CERT='-----BEGIN CERTIFICATE-----
-----END CERTIFICATE-----
'

→ om --env env.yml bosh-env --bosh --credhub
export BOSH_CLIENT_SECRET=secret
export BOSH_ENVIRONMENT=10.0.0.10
export BOSH_CA_CERT='-----BEGIN CERTIFICATE-----
-----END CERTIFICATE-----
'
export CREDHUB_CLIENT=client
export CREDHUB_SECRET=secret
export CREDHUB_SERVER=https://10.0.0.10:8844
export CREDHUB_CA_CERT='-----BEGIN CERTIFICATE-----
-----END CERTIFICATE-----
'
export BOSH_CLIENT=client
```

kcboyle:
The implementation looks good to me. Simple additions that slid in neatly with the existing code. 

Bonus points for regenerating the docs. I'll add this into the changelog

PR #524, opened 2020-11-30
configure-opsman: allow updating the UAA tokens expiration

jvshahid:
This PR allows users to update the UAA tokens expiration using the `configure-opsman` subcommand.  This change is useful for environments with strict compliance requirements such that the default access token and refresh token expiration are too long.

### DIFF
--- a/api/configure_opsman_service.go
+++ b/api/configure_opsman_service.go
@@ -35,6 +35,12 @@ type PivnetSettings struct {
 	APIToken string `json:"api_token" yaml:"api_token"`
 }
 
+type TokensExpiration struct {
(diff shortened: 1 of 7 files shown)


jtarchie:
Tested out locally. Applied the updated values as `10` and the subsequent `om curl` got it back.

```
go run main.go --env ~/workspace/pivotal/platform-automation-deployments/reference-gcp/env.yml curl --path /api/v0/uaa/tokens_expiration
Status: 200 OK

{
  "tokens_expiration": {
    "access_token_expiration": 10,
    "refresh_token_expiration": 10,
    "session_idle_timeout": 10
  }
}
```

Issue #533, opened 2021-01-12
Make om curl a bit more like regular curl


jghiloni:
There are a couple of small changes that could be made to `om curl` that would make them a bit more `curl`-ish. Specifically,

1. When `-d` is passed, `-x` should default to `POST`, not `GET`.
2. Allow users to pass files to `-d`, like curl's `-d @path-to-file`, which would slurp the contents of the file and pass that.

If these are interesting, I will investigate adding.

anEXPer:
I used the first of these two items for an interview question and came up with a decent implementation, which I'll PR presently.

I'm not sure there isn't a better way to check if the request flag is being explicitly passed or not, but I'm satisfied with the kind of bad way we came up with as a first draft.

anEXPer:
It turns out our implementation was on Ancient Code that I happened to have checked out, and ran afoul of some nice refactoring that I didn't want to untangle. So I've withdrawn the PR and may take another crack at it in the future.

jaristiz:
As there is a PR solving this issue, will meek it and close it once we have reviewed the PR.

PR #541, opened 2021-03-22
Improve curl

anEXPer:
This partially addresses issue #533.

I'm not sure there isn't a better way to check if the request flag is being explicitly passed or not, but I'm satisfied with the kind of bad way we came up with as a first draft; I'd be super appreciative of code review that addresses this part in particular - is there a way to do this in jhonda?

### DIFF
--- a/commands/curl.go
+++ b/commands/curl.go
@@ -53,7 +53,7 @@ func (c Curl) Execute(args []string) error {
 		Headers: requestHeaders,
 	}
 
-	if c.Options.Data != "" && c.Options.Method == "GET" {
+	if c.Options.Data != "" && c.requestNotSpecified(args) {
 		input.Method = "POST"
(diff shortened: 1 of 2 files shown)


jtarchie:
Do a `git pull --rebase` @anEXPer. This was done on a much older version of `om`.

anEXPer:
JT, nice to see you!

No _kidding_ is it an old one. And there's a major refactor that impacts this area of functionality, so rebasing is hard and simply cherry-picking the stuff onto main is incoherent/fails the tests now.

Well, live and learn, maybe I'll re-implement it on the current code as my next interview. I'm going to close this PR now, 'cause I don't actually have time to figure out why it's broken right now. I'm going to leave the branch for a bit, just in case I gain that time in the future.

Thank you!

Issue #558, opened 2021-09-02
`Revert` link on Ops Manager UI behaves unexpectedly for BOSH director tile


yosepkim:
After getting the tile settings using OM CLI (ex. `om -e env.yml staged-config -p p-healthwatch -c >p-healthwatch.yml`), pushing the same file to the OpsManager API (ex. `om --env env/env.yml configure-product --config p-healthwatch.yml`), this would show the `Revert` link only when applying changes to the BOSH director.

It is due to subnets' GUID not being persisted between API calls. Here are steps to reproduce:

```
om --env env.yml --trace staged-director-config --no-redact > director-config.yml
om --env env.yml --trace configure-director --config director-config.yml
```

Checking the API calls made you would notice that the guid subnets property is not set in the update API call.

```
GET /api/v0/staged/director/networks HTTP/1.1
Content-Type: application/json


HTTP/1.1 200 OK
Transfer-Encoding: chunked
Cache-Control: no-cache, no-store
Connection: keep-alive
Content-Type: application/json; charset=utf-8
Date: Mon, 16 Aug 2021 08:22:45 GMT
Etag: W/"ef0611ed04a22b673223b1d91badd7ae"
Expires: Fri, 01 Jan 1990 00:00:00 GMT
Pragma: no-cache
Referrer-Policy: strict-origin-when-cross-origin
Server: Ops Manager
Strict-Transport-Security: max-age=31536000; includeSubDomains
X-Content-Type-Options: nosniff
X-Download-Options: noopen
X-Frame-Options: SAMEORIGIN
X-Permitted-Cross-Domain-Policies: none
X-Request-Id: cf3fdfcb-bdd4-4a3a-bdf7-0ebe3404097e
X-Runtime: 0.207417
X-Xss-Protection: 1; mode=block

2bf
[...]

PR #559, opened 2021-09-04
Associate existing subnet GUIDs when configuring director networks

RageZBla:
Ref #558

### DIFF
--- a/api/director_service.go
+++ b/api/director_service.go
@@ -25,8 +25,15 @@ type Networks struct {
 }
 
 type Network struct {
+	GUID    string                 `yaml:"guid,omitempty"`
+	Name    string                 `yaml:"name"`
+	Subnets []Subnet               `yaml:"subnets,omitempty"`
+	Fields  map[string]interface{} `yaml:",inline"`
+}
+
+type Subnet struct {
 	GUID   string                 `yaml:"guid,omitempty"`
-	Name   string                 `yaml:"name"`
+	CIDR   string                 `yaml:"cidr"`
 	Fields map[string]interface{} `yaml:",inline"`
 }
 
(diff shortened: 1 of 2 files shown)


api/director_service.go
```diff
@@ -262,6 +270,19 @@ func (a Api) addGUIDToExistingNetworks(networks Networks) (Networks, error) {
 	return networks, nil
 }
 
+func (a Api) addGUIDToExistingSubnet(network *Network, existingNetwork *Network) []Subnet {
+	for k, subnet := range network.Subnets {
+		for _, existingSubnet := range existingNetwork.Subnets {
+			if subnet.CIDR == existingSubnet.CIDR {
```
RageZBla, line 276: Let me know if the comparison should be made on more fields. In theory it does not really matter since the other field would be updated.

wyike:
In my case, when the az configuration is:
```
az-configuration:
  - name: AZ1
    iaas_configuration_name: default
    clusters:
      - cluster: os-compute-cluster-1
        drs_rule: MUST
        guid: 5a132e367e2a82348a21
        host_group: null
        resource_pool: PKS
      - cluster: os-compute-cluster-1
        drs_rule: MUST
        guid: d59e16f5431b8beeb615
        host_group: null
        resource_pool: PKSWorkerPool
    guid: 54e4cdfc17edc55ae5d

```
and will meet error when staging back the director configuration:
```
cc
{"errors":{"base":["Cannot delete the cluster 'os-compute-cluster-1' in the availability zone 'AZ1' of a deployed product","Cannot have multiple clusters with the same guid. Guid: 5a132e367e2a82348a21"]}}
0
```

This change doesn't work to fix the problem. But I can use this change to simple fix my problem: https://gitlab.eng.vmware.com/PKS/gear2/-/commit/28838ed57a3edf699ea1e619c0bc45d363df00a2, not sure if you guys can open it.

xtremerui:
@dtimm @jaristiz FYI ^^

rhall-pivotal:
Closing due to age. If this is still needed, please resubmit. 
-TAS Operability

PR #560, opened 2021-09-14
fixes #554; don't fail if both gcp service account are set

### LINKED ISSUE #554: Why is it forbidden to set both `gcp_service_account` and `gcp_service_account_name`?
Hi,

why is it forbidden to set both `gcp_service_account` and `gcp_service_account_name` (https://github.com/pivotal-cf/om/blob/main/vmlifecycle/vmmanagers/gcp.go#L121)?

As far as i understand it the `gcp_service_account` are the credentials (json) used to create the opsman VM itself and the `gcp_service_account_name` is the email address of the service account that will be associated with the VM (https://cloud.google.com/compute/docs/access/create-enable-service-accounts-for-instances#using).
Those can and probably should be different service accounts with different IAM permissions (least privilege).
The SA associated with the Opsman VM can be used by the BOSH director as described here: https://docs.pivotal.io/ops-manager/2-10/gcp/config-manual.html#gcp-config
This has the advantage that one does not need to create a service account key for the vm associated SA as all authentication is handled GCP internally.
The documentation is unfortunately not very precise about the fact that one could use different service accounts.
https://docs.pivotal.io/ops-manager/2-10/gcp/prepare-env-manual.html#iam_account
https://docs.pivotal.io/ops-manager/2-10/gcp/deploy-manual.html#start-vm

The code seems to be just fine with both being set.
https://github.com/pivotal-cf/om/blob/main/vmlifecycle/vmmanagers/gcp.go#L164
https://github.com/pivotal-cf/om/blob/main/vmlifecycle/vmmanagers/gcp.go#L288
It uses the `gcp_service_account` to authenticate with GCP to create the vm, image, etc. and the `gcp_service_account_name` gets associated with the VM.

I suggest to remove the check if both are set.

Best Regards
Martin

martinpfeifer:


alejandra-lara (approved):
Hi, I consider that your changes are good

PR #562, opened 2021-09-16
Update tests for variable contentlength

wayneadams:
The result of `form.AddFile("file1", fileWithContent1)` seems to indeterminate. To account for this, we're adding the other possible state that shows up in the tests.


jaristiz (commented):
Seems OK to me.

PR #563, opened 2021-09-16
bugfix: PLATAUTO-4 Error while parsing a string with a double quote

jaristiz:
This PR fixes a problem while parsing a double quote from a config file.

Link: PLATAUTO-4


wayneadams (approved):
solid, release the commits!

PR #565, opened 2021-10-12
Update docs to contain new changes to commands

iplay88keys:
There are multiple commands which have command help output that is different than what is in the `docs` folder. I ran `go run docsgenerator/update-docs.go` and this PR contains the updated documentation.


jaristiz:
Looks OK to me.

Issue #573, opened 2022-01-05
om fails to authenticate if password has double quotes in it.


dforbes-pivotal:
Using om version 5.0.19 it has been noticed that using double quotes is resulting in a failed authentication.  Escaping the character is resulting in success.  

For example:=
```
root@4f12ff08-3b58-4e13-4426-f2cff59398de:/tmp/build/35bf6b7a# env  | grep -i mu3
OM_VAR_om_decryption_pwd=+qh4CCbaMu3"
root@4f12ff08-3b58-4e13-4426-f2cff59398de:/tmp/build/35bf6b7a# om --env env/vars/dev02/env.yml configure-director --config config/configs/director/director-2.10.24.yml --vars-file config/vars/global/vars.yml --vars-file config/vars/global/secrets.yml --vars-file config/vars/env_global/vars.yml --vars-file config/vars/env_global/secrets.yml --vars-file config/vars/dev02/vars.yml --vars-file config/vars/dev02/secrets.yml --vars-file vars/input_versions/dev.yml
2021/12/22 14:07:34 could not list installations: could not make api request to installations endpoint: could not send api request to GET /api/v0/installations: could not unlock ops manager, check if the decryption passphrase is correct

root@4f12ff08-3b58-4e13-4426-f2cff59398de:/tmp/build/35bf6b7a# export OM_VAR_om_decryption_pwd=+qh4CCboMu3\\\"
root@4f12ff08-3b58-4e13-4426-f2cff59398de:/tmp/build/35bf6b7a# env  | grep -i mu3
OM_VAR_om_decryption_pwd=+qh4CCbaMu3\"
root@4f12ff08-3b58-4e13-4426-f2cff59398de:/tmp/build/35bf6b7a# om --env env/vars/dev02/env.yml configure-director --config config/configs/director/director-2.10.24.yml --vars-file config/vars/global/vars.yml --vars-file config/vars/global/secrets.yml --vars-file config/vars/env_global/vars.yml --vars-file config/vars/env_global/secrets.yml --vars-file config/vars/dev02/vars.yml --vars-file config/vars/dev02/secrets.yml --vars-file vars/input_versions/dev.yml
started configuring director options for bosh tile
finished configuring director options for bosh tile
started configuring availability zone options for bosh tile
successfully fetched AZs, continuing
finished configuring availability zone options for bosh tile
started configuring network options for bosh tile
finished configuring network options for bosh tile
started configuring network assignment options for bosh tile
[...]

Issue #572, opened 2021-12-27
configure-saml-authentication is failing for new opsman


goriparthivamsi:
Upgraded to latest version of OM CLI 7.4.1 and noticed that SAML authentication is failing for new opsmanager VM. 
Error is as follows: 
om --env env/env.yml --skip-ssl-validation configure-saml-authentication --config config/((redacted))/saml.yml
configuring SAML authentication...
2021/12/27 18:28:51 could not configure authentication: request failed: unexpected response from /api/v0/setup:
HTTP/1.1 422 Unprocessable Entity
Transfer-Encoding: chunked
Cache-Control: no-cache, no-store
Connection: keep-alive
Content-Type: application/json; charset=utf-8
Date: Mon, 27 Dec 2021 18:28:51 GMT
Expires: Fri, 01 Jan 1990 00:00:00 GMT
Pragma: no-cache
Referrer-Policy: strict-origin-when-cross-origin
Server: Ops Manager
Strict-Transport-Security: max-age=31536000; includeSubDomains
X-Content-Type-Options: nosniff
X-Download-Options: noopen
X-Frame-Options: SAMEORIGIN
X-Permitted-Cross-Domain-Policies: none
X-Request-Id: 83d134ae-de9f-41bb-864f-d384cf0748fb
X-Runtime: 0.084294
X-Xss-Protection: 1; mode=block

34
{"errors":{"setup":["IDP Metadata is invalid XML"]}}
0

Switched to older version of CLI which is 7.3.0 and everything is working as expected. I think the issue has been introduced in the latest versions.

scottgai:
My one customer also hit the issue after upgrading to latest version of OM CLI 7.4.1 and noticed that SAML authentication is failing for Ops Manager VM.
Error is as follows:
`om --env env/env.yml --skip-ssl-validation configure-saml-authentication --config config/((redacted))/saml.yml
configuring SAML authentication...
2021/12/27 18:28:51 could not configure authentication: request failed: unexpected response from /api/v0/setup:
HTTP/1.1 422 Unprocessable Entity
Transfer-Encoding: chunked
Cache-Control: no-cache, no-store
Connection: keep-alive
Content-Type: application/json; charset=utf-8
Date: Mon, 27 Dec 2021 18:28:51 GMT
Expires: Fri, 01 Jan 1990 00:00:00 GMT
Pragma: no-cache
Referrer-Policy: strict-origin-when-cross-origin
Server: Ops Manager
[...]

PR #576, opened 2022-01-07
Removed double quotes escaping

jaristiz:
Removed escaping of double-quotes on parameter values which caused
problems when parsing correctly formed parameters.

Authored-by: Jhonathan Aristizabal <jhonathana@vmware.com>


jaristiz:
This solves #573 and #572

PR #579, opened 2022-01-10
import-installation: Handle bad gateway responses like connection refused

dtimm:
per https://www.pivotaltracker.com/story/show/179912729


crhntr (changes requested):
From a readability perspective, it might be nice to pull this condition into a separate helper function something like:

```go
func isErrThatMightResolveOnRetry(err error) bool { return false /* copy condition here */ }
```

You could then unit test the helper function and reduce the command level test burden.

crhntr (approved):
Looks good

PR #582, opened 2022-02-04
Add darwin_arm64 builds to goreleaser

dtimm:
This will add automatic builds for Apple computers with M1 architecture to published releases.

It changes the names of build artifacts to include architecture, as well.


jtarchie:
Changing the names of the binaries will potentially ruin downstream consumers if they are using github releases resource in Concourse. Adding that to the `CHANGELOG` would be helpful.

PR #591, opened 2022-10-24
added support for jammy stemcell

jaristiz:
- Added support for jammy stemcell on the `stow-client`

Authored-by: Jhonathan Aristizabal <jhonathana@vmware.com>


anEXPer (approved):
This is the correct change, thanks!

PR #636, opened 2024-04-04
Update README.md

joyvuu-dave:
GO112MODULE is not a recognized variable in golang. The correct name is GO111MODULE.


wayneadams:
Thans for the catch, @joyvuu-dave !

Issue #601, opened 2023-05-20
Update configure-ldap-authentication command to have an option to set ldap_max_search_depth


mstaplestanzutam:
### Is your feature request related to a problem? Please describe.

ldap_max_search_depth is required as part of TAS 3.0, and while OpsMan will be updated to provide this with a value of 1 if this is not set, it would be nice to be able to set this value via OM so it can be included in automation. 

### Describe the solution you'd like

Inclusion of the ability to change ldap_max_search_depth in OM. 

### Describe alternatives you've considered

Operations Manager will be updated to ensure a value of 1 is added if their is no value set, but we have some use cases where being able to set this for automation purposes would be nice, especially if a value needs to be set.

rhall-pivotal:
Merged https://github.com/pivotal-cf/om/pull/649

PR #649, opened 2024-05-22
Add optional --ldap-max-search-depth to fix issue #601

sneal:
Added an optional `--ldap-max-search-depth` flag to use with Operations Manager 3.0+. The new flag can be used like so:

```
om configure-ldap-authentication \
  --decryption-passphrase='supersecret' \
  --email-attribute='email' \
  --group-search-base='ou=Groups,dc=vmware,dc=com' \
  --group-search-filter='Corp' \
  --ldap-password='supersecret' \
  --ldap-rbac-admin-group-name='Admins' \
  --ldap-referrals=follow \
  --ldap-username=ldapadmin \
  --ldap-max-search-depth=5 \
  --server-ssl-cert="$(cat ~/ldap.crt)" \
  --server-url='ldaps://ad.vmware.com' \
  --user-search-base='Users' \
  --user-search-filter='devs' \
  --skip-create-bosh-admin-client \
  --precreated-client-secret='supersecret'
  ```

### DIFF
--- a/api/setup_service.go
+++ b/api/setup_service.go
@@ -17,6 +17,7 @@ type LDAPSettings struct {
 	LDAPRBACAdminGroup string `json:"ldap_rbac_admin_group_name,omitempty"`
 	LDAPReferral       string `json:"ldap_referrals,omitempty"`
 	LDAPUsername       string `json:"ldap_username,omitempty"`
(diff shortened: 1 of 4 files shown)


rhall-pivotal:
Pulled in the changes and rebased on main. Tests ran locally without issue. Merging!

PR #661, opened 2024-07-31
bump ginkgo to v2.19.x

yharish991:
The current version of ginkgo being used in the project is no longer supported and needs to be bumped to latest version to prevent technical debt and to get patches for security vulnerabilities


CONTRIBUTING.md
```diff
@@ -51,7 +51,7 @@ docker run \
   -v $(pwd):/workspace \
   -w /workspace internalpcfplatformautomation/ci:testing \
   /bin/bash \
-  -c "go install github.com/onsi/ginkgo/ginkgo && CGO_ENABLED=1 ginkgo -r -race -succinct -randomizeAllSpecs -randomizeSuites -keepGoing -skipPackage acceptance,vmlifecyclecommands,vmmanagers"
+  -c "go install github.com/onsi/ginkgo/v2 && CGO_ENABLED=1 ginkgo -r -race -succinct -randomizeAllSpecs -randomizeSuites -keepGoing -skipPackage acceptance,vmlifecyclecommands,vmmanagers"
```
davewalter, line 54: According to the [docs](https://onsi.github.io/ginkgo/), this should be `go install github.com/onsi/ginkgo/v2/ginkgo`.

depstability/records/depnames-6.1.2.txt
```diff
@@ -81,7 +81,7 @@ github.com/ncw/swift
 github.com/nu7hatch/gouuid
 github.com/nxadm/tail
 github.com/olekukonko/tablewriter
-github.com/onsi/ginkgo
+github.com/onsi/ginkgo/v2
```
davewalter, line 84: Should these depstability records be changed? I'm not familiar with depstability, but it feels like these should be historical records of the dependencies for specific versions. The changes to 7.11.0 and 7.12.0 seem especially strange.

commands/import_installation_test.go
```diff
@@ -192,7 +194,7 @@ var _ = Describe("ImportInstallation", func() {
 			Expect(fmt.Sprintf(format, v...)).To(Equal("finished import"))
 		})
 
-		It("it only retries 3 times before giving up", func(done Done) {
+		It("it only retries 3 times before giving up", func(ctx context.Context) {
```
davewalter, line 197: Is the Context in the test function necessary? It doesn't seem to be used.

vmlifecycle/vmlifecyclecommands/upgrade_opsman_test.go
```diff
@@ -454,43 +457,43 @@ var _ = Describe("upgradeOpsman", func() {
...
+							Entry("semver with build vs build, via minor", "2.4-build.1", "2.5.1-build.103"),
+							Entry("semver with build vs build, via major", "1.5-build.1", "2.5.1-build.103"),
+							Entry("build vs semver with build, via patch", "2.5.0-build.103", "2.5-build.1"),
+							Entry("build vs semver with build, via minor", "2.4.1-build.103", "2.5-build.1"),
+							Entry("build vs semver with build, via major", "1.5.1-build.103", "2.5-build.1"),
 						)
 					}
 
-					It("polls till the provided timeout then errors", func(done Done) {
+					It("polls till the provided timeout then errors", func(ctx context.Context) {
```
davewalter, line 496: Is the Context in the test function necessary? It doesn't seem to be used.

commands/import_installation_test.go
```diff
@@ -265,7 +266,7 @@ var _ = Describe("ImportInstallation", func() {
...
```
davewalter, line 269: Is the Context in the test function necessary? It doesn't seem to be used.

davewalter (commented):
Looks good overall. Please see the inline comments for a few minor things that I think need to be addressed.

CONTRIBUTING.md
yharish991: thanks for catching this

commands/import_installation_test.go
yharish991: that's the way to test async behavior in ginkgo, without the context being passed in the callback, we get this error 
```
Invalid NodeTimeout SpecTimeout, or GracePeriod
It("it only retries 3 times before giving up", func() {
/Users/hyayi/work/om/commands/import_installation_test.go:197
  [It] was passed NodeTimeout, SpecTimeout, or GracePeriod but does not have a
  callback that accepts a SpecContext or context.Context.  You must accept a
  context to enable timeouts and grace periods

  Learn more at:
  http://onsi.github.io/ginkgo/#spec-timeouts-and-interruptible-nodes

```

vmlifecycle/vmlifecyclecommands/upgrade_opsman_test.go
yharish991: explained in the [previous](https://github.com/pivotal-cf/om/pull/661#discussion_r1698970284) comment

commands/import_installation_test.go
yharish991: explained in the [previous](https://github.com/pivotal-cf/om/pull/661#discussion_r1698970284) comment

depstability/records/depnames-6.1.2.txt
yharish991: hmm. I'm not sure why the other versions got updated, let me dig more into this, all i did was run the test and looks like that updated the files

yharish991: fixed this issue and pushed the change

commands/import_installation_test.go
davewalter: I was expecting to see the Context being used to actually terminate the test when it times out (as in the example in the docs). It seems like this test would ignore the time out and continue to run until completion.

yharish991: NodeTimeout/SpecTimeout terminates the spec, I think that should be enough for this test. However, I refactored this test to use context with a timeout, which should also timeout the process the test triggered. Let me know what you think

davewalter (approved):

PR #663, opened 2024-08-05
Replace deprecated methods usage with alternative methods

yharish991:
This PR replaces all Go standard library deprecated methods uswith the recommended alternatives.

### DIFF
--- a/api/staged_products_service_test.go
+++ b/api/staged_products_service_test.go
@@ -682,7 +682,7 @@ valid options configurations include percentages ('50%'), counts ('2'), and 'def
 		//		if strings.Contains(req.URL.Path, "some-product-guid") {
 		//			resp = &http.Response{
 		//				StatusCode: http.StatusOK,
-		//				Body: ioutil.NopCloser(bytes.NewBufferString(`{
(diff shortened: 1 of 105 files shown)


api/staged_products_service_test.go
```diff
@@ -682,7 +682,7 @@ valid options configurations include percentages ('50%'), counts ('2'), and 'def
 		//		if strings.Contains(req.URL.Path, "some-product-guid") {
 		//			resp = &http.Response{
 		//				StatusCode: http.StatusOK,
-		//				Body: ioutil.NopCloser(bytes.NewBufferString(`{
+		//				Body: io.NopCloser(bytes.NewBufferString(`{
```
davewalter, line 685: This code is commented out. Can it be removed?

davewalter (approved):
Looks good. The only question was about commented-out code, but nothing that blocks this from being merged.

api/staged_products_service_test.go
yharish991: I did see this, but I was not exactly sure what this BeforeEach block, so left it as is. Will address this in a followup PR as this is not in the scope of this PR

PR #678, opened 2024-11-25
Sends helpful error when metedata has multiple stemcell types

wayneadams:
- Add checkForSingleStemcellType function to validate stemcell selection
- Update getLatestStemcell to track all unique stemcell types
- Add test case to verify error handling for multiple stemcell types

Authored by: Wayne Adams <wayne.adams@broadcom.com>

### DIFF
--- a/download_clients/pivnet_client.go
+++ b/download_clients/pivnet_client.go
@@ -170,25 +170,47 @@ func (p *pivnetClient) checkForSingleProductFile(glob string, productFiles []piv
 	return nil
 }
 
+func (p *pivnetClient) checkForSingleStemcellType(stemcellSlugs []string) error {
+    if len(stemcellSlugs) > 1 {
+        return fmt.Errorf("multiple stemcell types found: %s", strings.Join(stemcellSlugs, ", "))
+    }
+    return nil
+}
+
 func (p *pivnetClient) getLatestStemcell(dependencies []pivnet.ReleaseDependency) (string, string, error) {
 	var (
-		stemcellSlug string
+		stemcellSlugs []string
 		versions     []string
 	)
 
 	for _, dependency := range dependencies {
 		if strings.Contains(dependency.Release.Product.Slug, "stemcells") {
-			stemcellSlug = dependency.Release.Product.Slug
+			if !contains(stemcellSlugs, dependency.Release.Product.Slug) {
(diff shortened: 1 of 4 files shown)


download_clients/pivnet_client.go
```diff
@@ -170,25 +170,47 @@ func (p *pivnetClient) checkForSingleProductFile(glob string, productFiles []piv
...
 	stemcellVersion, err := getLatestStemcellVersion(versions)
 	if err != nil {
 		return "", "", err
 	}
 
-	return stemcellSlug, stemcellVersion, nil
+	return stemcellSlugs[0], stemcellVersion, nil
+}
+
+func contains(slice []string, str string) bool {
```
drich10, line 207: Didn't pull the branch down, but I think you can use the `slices` package to do contains operations instead of the helper function.

https://freshman.tech/snippets/go/check-if-slice-contains-element/

drich10 (changes requested):
Looks good to me - minor comment for using some of the "newer" go functionality.

Food for thought, that isn't in scope of this PR. Would it make sense to extend _choice_ in this scenario so that folks could pick an OS/slug as part of invoking the om command?

download_clients/pivnet_client.go
wayneadams: Good call. I've added the newness

download_clients/pivnet_client.go
```diff
@@ -170,25 +172,38 @@ func (p *pivnetClient) checkForSingleProductFile(glob string, productFiles []piv
...
 		if strings.Contains(dependency.Release.Product.Slug, "stemcells") {
-			stemcellSlug = dependency.Release.Product.Slug
+			if !slices.Contains(stemcellSlugs, dependency.Release.Product.Slug) {
+				stemcellSlugs = append(stemcellSlugs, dependency.Release.Product.Slug)
+			}
 			versions = append(versions, dependency.Release.Version)
 		}
 	}
 
+	if err := p.checkForSingleStemcellType(stemcellSlugs); err != nil {
```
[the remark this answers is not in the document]
wayneadams: I actually had it inline first, but then noticed there was already a similar function here: https://github.com/pivotal-cf/om/pull/678/files#diff-d8151afadbdb2454bb3ab3287e73f1e44c90f777f1d3ab757d59e2078512ca20R161. So it seems appropriate to follow that expressiveness?

yharish991 (approved):

PR #680, opened 2024-12-03
vmlifecycle - aws - add support for setting volume type

rabeyta:
default to gp2 (current behavior). allows user to define boot_disk_type in their aws config to instruct the type of volume to use when modifying the volume after vm creation

refactoring aws lifecycle create to no longer perform a modify volume after initial creation. pass those disk options to initial creation to allow users to modify the volume if necessary after initial create without a penalty. aws imposes a 6 hour cool down between changes and this allows a user to modify vs the create causing a 6 hour cooldown before the next change is allowed


krsri:
LGTM

PR #681, opened 2024-12-17
Update bosh-cli go module to v7

rajathagasthya:
bosh-cli recently updated bosh-agent to v2 and to get that bosh-cli needs to be updated to v7 in this repo.


rajathagasthya:
@wayneadams I'm not sure if I'm supposed to create a new `depnames-*.txt` file since this PR is changing dependency. Are you able to advise what to do?

PR #682, opened 2025-02-28
add stemcell-slug option to download-product command

yharish991:
This PR adds a `stemcell-slug` optinal option to `download-product` command to specify the stemcell slug in case of multiple stemcells


download_clients/pivnet_client_test.go
wayneadams, line 1: These are great new tests!

wayneadams (approved):
LGTM

drich10 (changes requested):
Overall looks good! Just a few changes would make the new logic a bit easier to parse for our future selves

drich10 (approved):
thanks for the updates and lgtm!

PR #685, opened 2025-04-21
Ensure existing clusters within availability zones match correctly

drich10:
The previous implementation expected cluster names to be unique. Unfortunately, this is not true across infras so we instead assume the configuration is complete for a given cluster if the guid is already set.


yharish991 (approved):
LGTM!!

PR #690, opened 2025-05-30
fix(cli): fail fast on unknown config flags

sanjimoh:
Previously, the OM CLI would silently ignore unknown arguments when parsing config-related flags due to the use of `flags.IgnoreUnknown` in loadConfigFile. This could lead to user mistakes going unnoticed.

This change removes `flags.IgnoreUnknown` and uses strict parsing for config flags, ensuring that any unknown or misspelled config-related arguments cause the CLI to fail fast with a helpful error message.

This improves user feedback and prevents misconfiguration due to typos or unsupported flags.

### DIFF
--- a/cmd/main.go
+++ b/cmd/main.go
@@ -848,9 +848,12 @@ func validateCommandFlags(parser *flags.Parser, args []string) error {
 	invalidFlags := findUnknownFlags(selectedCmd, args[currentArgPosition:])
 
 	if len(invalidFlags) > 0 {
-		fmt.Fprintf(os.Stderr, "Error: unknown flag(s) %q for command '%s'\n", invalidFlags, selectedCmd.Name)
-		fmt.Fprintf(os.Stderr, "See 'om %s --help' for available options.\n", selectedCmd.Name)
-		os.Exit(1)
+		return fmt.Errorf(
+			"Error: unknown flag(s) %q for command '%s'\nSee 'om %s --help' for available options",
+			invalidFlags,
+			selectedCmd.Name,
+			selectedCmd.Name,
+		)
 	}
 	return nil
 }


drich10:
With the command `om -e om.yml upload-product -p tiles-target/cf-6.0.5-build.2.pivotal -r 18000 -o 18000`, where `-r` and `-o` are not valid command options (but are valid application options) - I expect a failure complaining one or both of those options are invalid for the command. The current changeset instead incorrectly complains that `-p` is an invalid option for the command. 

image

sanjimoh:
> With the command `om -e om.yml upload-product -p tiles-target/cf-6.0.5-build.2.pivotal -r 18000 -o 18000`, where `-r` and `-o` are not valid command options (but are valid application options) - I expect a failure complaining one or both of those options are invalid for the command. The current changeset instead incorrectly complains that `-p` is an invalid option for the command.

@drich10, it was a misunderstanding on my part regarding the problem. Thank you for helping me understand the problem statement. I’ve pushed a new changeset with the intention of fixing this issue now. 
```
$ om -e $OM_ENV upload-product -p tas_hub_collector-10.1.1747720947.pivotal -r 18000 -o 18000
Error: unknown flag(s) ["-r" "-o"] for command 'upload-product'
See 'om upload-product --help' for available options.
unknown flag(s) ["-r" "-o"] for command 'upload-product'
```

Could you please review this new changeset?

main.go
```diff
@@ -18,9 +18,10 @@ func main() {
 	err := cmd.Main(os.Stdout, os.Stderr, version, applySleepDurationString, os.Args)
 	if err != nil {
 		if errors.Is(err, commands.ErrBoshDiffChangesExist) {
-			log.Print(err)
+			fmt.Fprintln(os.Stderr, err)
```
jtarchie, line 21: I'm coming from `om` maintainers past.

The context for the use of `log` was to ensure timestamps. Timestamps are always important in the automation flow.

cmd/validate_command_flags_test.go
```diff
@@ -0,0 +1,50 @@
+package cmd
```
drich10, line 1: I think this test would be better suited to the acceptance suite rather than expanding the tests that exist in the `cmd` pacakge. I'm not quite sure what the representative set of commands and their permutations are, but I'd like us to test a handful of the actual commands rather than recreating and unit testing them

drich10: I think the loadConfigFile_test was a bit of a misdirect on what convention to follow here.

drich10 (changes requested):
Overall it looks pretty good! Two things:
- I'd just like to see the test moved to the acceptance suite and tie the tests to some of the existing commands to tie the tests to the currently supported interface
- Can you explain the reasoning for the stderr printing on top of the error return? Its a departure from the output that occurs currently when a flag is incorrectly specified.

sanjimoh:
> Overall it looks pretty good! Two things:
> 
> * I'd just like to see the test moved to the acceptance suite and tie the tests to some of the existing commands to tie the tests to the currently supported interface
> * Can you explain the reasoning for the stderr printing on top of the error return? Its a departure from the output that occurs currently when a flag is incorrectly specified.

Thanks @drich10 for the review! I've addressed both the review comments now. PTAL & let me know if it looks good to merge now.

PR #691, opened 2025-06-04
fix: prevent duplicate stemcell uploads

sanjimoh:
Previously, om would re-upload a stemcell even if it was already present in Ops Manager, due to mismatches in filename conventions & incomplete checks. This caused unnecessary uploads & pipeline delays.

This PR addresses the issue by -
- Extracting the OS, version, & infrastructure (IAAS) directly from the stemcell manifest (stemcell.MF) instead of relying on the filename.
- Matching these fields against the diagnostic report's available stemcells & infrastructure_type, ensuring a stemcell is only uploaded if it is truly absent for the given OS, version, & IAAS.

**Alternatives considered:**
- Parsing IAAS from the stemcell filename in the diagnostic report (error-prone & brittle).

**Final solution:**
- Use the manifest's `operating_system`, `version`, & `cloud_properties.infrastructure` fields.
- Compare these to the diagnostic report's `os`, `version`, and `infrastructure_type` fields for robust, intent-driven matching.

Also includes new & updated tests to verify the improved logic and prevent regressions.

### DIFF
--- a/acceptance/download_product_pivnet_test.go
+++ b/acceptance/download_product_pivnet_test.go
@@ -13,6 +13,9 @@ import (
 	"strings"
 	"time"
 
+	"archive/tar"
+	"compress/gzip"
+
 	"cloud.google.com/go/storage"
 	. "github.com/onsi/ginkgo/v2"
 	. "github.com/onsi/gomega"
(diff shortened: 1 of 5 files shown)


jtarchie:
For historical context, the reason was for consistency. We aimed for reproducibility, which meant constantly uploading the stem cells. Since we always knew it would upload, there was never a case where it didn't upload. It was not a bug, it was very intentional.

I'm also no longer on the team or at Broadcom. I'm just providing some notes from 5+ years ago `om` team.

sankar17:
@jtarchie @sanjimoh , Since this PR created long ago

I have created new PR : https://github.com/pivotal-cf/om/pull/728
Manually verified and upload-stemcells is working as expected.

PR #696, opened 2025-08-25
fix: add wait till the vm started in aws create vm to avoid assignIP failures

ramramu3433:


ramramu3433:
Added tests for new running check and used direct match running instead of consists of

yharish991-brcm (approved):
Changes lgtm 👍 but I would wait for a review from @drich10

PR #701, opened 2025-09-05
Support multiple tags

krsri:
The --tags argument for az vm create expects each tag to be passed as a separate argument (e.g., --tags key1=val1 key2=val2).
In the earlier implementation, the entire azure.Tags string was appended as a single argument after --tags. As a result, Azure CLI interpreted all tags as one long string, causing only a single tag to be applied to the VM instead of multiple tags.


drich10 (approved):
Agreed w/ Rajan's comment - only recommendation is that we make sure this makes it into a _minor_ release given the change type as opposed to a patch

PR #702, opened 2025-09-15
fix: wait for AWS OpsManager VM to be running

blgm:
PR #696 added a wait for the AWS OpsManager VM to start running. This builds on top of that change, updating the expected output from the AWS CLI so that it works.

Tests were also updated. Note that the tests are quite fragile and would benefit from re-work. Specifically the
`runner.ExecuteWithEnvVarsReturnsOnCall()` test fake is problematic because (a) it's possible to write tests that pass but do not match the actual implmentation of the AWS CLI, and (b) adding/removing/changing the order of calls requires the tests to be re-written.


krsri (approved):
LGTM

PR #787, opened 2026-07-06
Fix stemcell availability check to fallback to filename matching

yharish991-brcm:
The CheckStemcellAvailability function was returning false immediately when manifest-based matching didn't find a stemcell, without attempting the filename-based fallback. This caused the download-product command with --check-already-uploaded to re-download stemcells that were already available in Ops Manager.

Remove the early return after manifest matching so the function can gracefully degrade to filename-based matching when manifest comparison doesn't find a match. This ensures stemcells are correctly detected regardless of whether the manifest fields match exactly.

Fixes TNZ-103785

### DIFF
--- a/api/stemcell_service.go
+++ b/api/stemcell_service.go
@@ -8,6 +8,8 @@ import (
 	"io"
 	"os"
 	"path/filepath"
+	"regexp"
+	"strings"
 
 	"gopkg.in/yaml.v2"
 )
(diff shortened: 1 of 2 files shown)


api/stemcell_service.go
```diff
@@ -111,16 +113,57 @@ func extractStemcellManifest(tgzPath string) (stemcellManifest, error) {
...
+// Handles both common formats:
+// - bosh-stemcell-1.1250-vsphere-esxi-ubuntu-jammy-go_agent.tgz
+// - bosh-vsphere-esxi-ubuntu-jammy-go_agent-1.1250.tgz
+func parseStemcellFilename(filename string) (os string, version string, infrastructure string) {
+	// Extract just the filename if a path is provided
+	filename = filepath.Base(filename)
+	filename = strings.TrimSuffix(filename, ".tgz")
+
+	// Try pattern 1: bosh-stemcell-VERSION-INFRASTRUCTURE-OS-go_agent
+	re1 := regexp.MustCompile(`^bosh-stemcell-(\d+\.\d+(?:\.\d+)?)-(.+?)-((?:ubuntu|centos|rhel|windows|alpine|debian).+?)-go_agent$`)
```
sankar17: @yharish991-brcm Customer reported issue with aws environment.  We need to make sure this fix works with aws setup. Also I see there is no pattern matching for aws image

api/stemcell_service.go
```diff
@@ -111,16 +113,57 @@ func extractStemcellManifest(tgzPath string) (stemcellManifest, error) {
...
+	if matches := re1.FindStringSubmatch(filename); matches != nil {
+		version = matches[1]
+		infrastructure = matches[2]
+		os = matches[3]
+		return
+	}
+
+	// Try pattern 2: bosh-INFRASTRUCTURE-OS-go_agent-VERSION
+	re2 := regexp.MustCompile(`^bosh-(.+?)-((?:ubuntu|centos|rhel|windows|alpine|debian).+?)-go_agent-(\d+\.\d+(?:\.\d+)?)$`)
```
sankar17, line 137: ditto

sankar17 (changes requested):

api/stemcell_service.go
yharish991-brcm: this will work for any infra `azure-hyperv`, `google-kvm`, `aws-xen-hvm`, `vsphere-esxi`

sankar17 (approved):

PR #822, opened 2026-09-06
Add unsafe-dependency bypass flags to om apply-changes

skaarthik:
Context:
Tiles can declare optional dependencies on other tiles. Ops Manager blocks apply-changes when an optional dependency has unsafe property changes on update, or when deleting a product whose optional dependent isn't marked safe-to-delete. Platform engineers had no way to proceed past these specific checks from the CLI even when they understood and accepted the risk — the only path was disabling verifiers more broadly via --ignore-warnings, which suppresses everything overridable rather than just this one class of check.

Implementation notes:
- Added --allow-unsafe-dependency-update and --allow-unsafe-dependency-deletion flags to `apply-changes`, wired through ApplyChanges.Options -> applyChangesService -> Api.CreateInstallation.
- Both default to false and are marshaled with `omitempty`, so they are only present in the POST /api/v0/installations body when set to true — the request is byte-for-byte unchanged for every caller that doesn't pass them.
- This is a CLI-only change. The Ops Manager backend support for these two params (permitted on the installations endpoint, wired through ValidationOverrideOptions into ProductDependencyVerifier) already exists as of Ops Manager releases/11.0 (introduced 2026-06-18); this branch only makes `om` use what's already there. Against an Ops Manager version predating that release, the fields are silently stripped by Rails strong-parameter filtering rather than causing a request error — the flags simply have no effect on those older targets.
- On a blocked apply-changes, Ops Manager's 422 response includes an `available_overrides` array naming which bypass(es) would unblock the specific failure. CreateInstallation now decodes that and appends a hint mapping known values to their CLI flag (falling back to the raw value for any override `om` doesn't recognize yet, per the API's forward-compatibility contract).
- Regenerated the counterfeiter fake for applyChangesService to match the new signature; regenerated docs/apply-changes/README.md via the repo's docsgenerator tooling rather than hand-editing it.

Usage notes:
- `om apply-changes --allow-unsafe-dependency-update`
- `om apply-changes --allow-unsafe-dependency-deletion`
- Both can be combined with each other and with existing flags (e.g. --product-name). Both are no-ops unless explicitly passed.
- If apply-changes is blocked and the failure has a known override, the resulting error now suggests the specific flag to retry with, instead of only pointing at `om pre-deploy-check`.
- Requires an Ops Manager release that includes this backend feature (11.0+); on older targets, passing these flags will not bypass anything and the original blocking behavior still applies.

### DIFF
--- a/commands/apply_changes.go
+++ b/commands/apply_changes.go
@@ -92,6 +92,16 @@ func (ac ApplyChanges) Execute(args []string) error {
 		changedProducts = ac.Options.ProductNames
 	}
 
+	if ac.Options.AllowUnsafeDependencyUpdate || ac.Options.AllowUnsafeDependencyDeletion {
+		info, err := ac.service.Info()
+		if err != nil {
+			return fmt.Errorf("could not retrieve info from targetted ops manager: %v", err)
+		}
+		if ok, err := info.VersionAtLeast(11, 0); !ok {
+			return fmt.Errorf("--allow-unsafe-dependency-update and --allow-unsafe-dependency-deletion are only available with Ops Manager 11.0 or later: you are running %s. Error: %w", info.Version, err)
+		}
+	}
+
 	installation, err := ac.service.RunningInstallation()
 	if err != nil {
 		return fmt.Errorf("could not check for any already running installation: %s", err)
--- a/commands/apply_changes.go
+++ b/commands/apply_changes.go
@@ -98,7 +98,10 @@ func (ac ApplyChanges) Execute(args []string) error {
 			return fmt.Errorf("could not retrieve info from targetted ops manager: %v", err)
 		}
 		if ok, err := info.VersionAtLeast(11, 0); !ok {
-			return fmt.Errorf("--allow-unsafe-dependency-update and --allow-unsafe-dependency-deletion are only available with Ops Manager 11.0 or later: you are running %s. Error: %w", info.Version, err)
+			if err != nil {
+				return fmt.Errorf("Could not determine Ops Manager version to accept flags --allow-unsafe-dependency-update / --allow-unsafe-dependency-deletion: %w", err)
+			}
+			return fmt.Errorf("--allow-unsafe-dependency-update and --allow-unsafe-dependency-deletion are only available with Ops Manager 11.0 or later: you are running %s", info.Version)
 		}
 	}
 
--- a/commands/apply_changes.go
+++ b/commands/apply_changes.go
@@ -20,19 +20,21 @@ type ApplyChanges struct {
 	logWriter      logWriter
 	waitDuration   time.Duration
 	Options        struct {
-		Config               string   `short:"c"   long:"config"               description:"path to yml file containing errand configuration (see docs/apply-changes/README.md for format)"`
-		IgnoreWarnings       bool     `short:"i"   long:"ignore-warnings"      description:"For convenience. Use other commands to disable particular verifiers if they are inappropriate."`
-		Reattach             bool     `long:"reattach" description:"reattach to an already running apply changes (if available)"`
-		RecreateVMs          bool     `long:"recreate-vms" description:"recreate all vms"`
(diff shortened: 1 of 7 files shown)


commands/apply_changes_test.go
```diff
@@ -131,19 +133,69 @@ var _ = Describe("ApplyChanges", func() {
...
+
 		When("passed the skip-deploy-products flag", func() {
 			It("applies changes while not deploying products", func() {
 				command := commands.NewApplyChanges(service, pendingService, writer, logger, 1)
 
 				err := executeCommand(command, []string{"--skip-deploy-products"})
 				Expect(err).ToNot(HaveOccurred())
 
-				_, _, deployProducts, _, _ := service.CreateInstallationArgsForCall(0)
+				_, _, deployProducts, _, _, _, _ := service.CreateInstallationArgsForCall(0)
```
sanjimoh: `deployProducts` is position 2, not 3. Should be `_, deployProducts, _, _, _, _, _`.

commands/apply_changes.go
```diff
@@ -158,8 +160,15 @@ func (ac ApplyChanges) Execute(args []string) error {
 		}
 	}
 
+	if ac.Options.AllowUnsafeDependencyUpdate {
+		ac.logger.Printf("allow-unsafe-dependency-update is set: unsafe optional-dependency update checks will be bypassed")
+	}
+	if ac.Options.AllowUnsafeDependencyDeletion {
+		ac.logger.Printf("allow-unsafe-dependency-deletion is set: unsafe optional-dependency deletion checks will be bypassed")
+	}
```
sanjimoh, line 181: `if` only checks that the flag was passed, not that the target Ops Manager can act on it. Per the PR description Ops Manager accepts these fields from 11.0 onward and silently drops them before that, so against an older Ops Manager om prints "checks will be bypassed" when nothing was, and the apply then fails with the same error the flag was meant to get past.

commands/apply_changes_test.go
skaarthik: Good catch! Addressed it.

commands/apply_changes.go
skaarthik: Added ops man version check to address this issue.

commands/apply_changes.go
```diff
@@ -90,6 +92,16 @@ func (ac ApplyChanges) Execute(args []string) error {
 		changedProducts = ac.Options.ProductNames
 	}
 
+	if ac.Options.AllowUnsafeDependencyUpdate || ac.Options.AllowUnsafeDependencyDeletion {
+		info, err := ac.service.Info()
+		if err != nil {
+			return fmt.Errorf("could not retrieve info from targetted ops manager: %v", err)
+		}
+		if ok, err := info.VersionAtLeast(11, 0); !ok {
+			return fmt.Errorf("--allow-unsafe-dependency-update and --allow-unsafe-dependency-deletion are only available with Ops Manager 11.0 or later: you are running %s. Error: %w", info.Version, err)
```
sanjimoh: `VersionAtLeast` returns `(false, nil)` when the version parses fine but is below the threshold, so `%w` gets a nil error and this prints `Error: %!w(<nil>)` in the normal case. Something like -
```
--allow-unsafe-dependency-update and --allow-unsafe-dependency-deletion are only available
with Ops Manager 11.0 or later: you are running 3.1.3. Error: %!w(<nil>)
```
Reproduced on 10.2-build1, 3.1.3 and 2.9-build.4; only a malformed version puts a real error there. Suggest splitting the two cases:
```
    if ok, err := info.VersionAtLeast(11, 0); !ok {
        if err != nil {
            return fmt.Errorf("could not determine Ops Manager version: %w", err)
        }
        return fmt.Errorf("--allow-unsafe-dependency-update and --allow-unsafe-dependency-deletion are only available with Ops Manager 11.0 or later: you are running %s", info.Version)
    }
```

skaarthik: Addressed this issue in the recent commit and added new unit tests

sanjimoh (approved):

PR #823, opened 2026-09-09
fix(TNZGOV-8041): remove temp vSphere options file after use

skaarthik:
## Summary

Fixes TNZGOV-8041 — vSphere `admin_password` persisted in undeleted temp options file (info-disclosure, MEDIUM).

`createOptionsFile()` in `vmlifecycle/vmmanagers/vsphere.go` writes the OVA import options — including the cleartext `admin_password` for legacy pre-2.6 `ssh_password` deployments — to a file created with `os.CreateTemp("", "options.json")`. `CreateVM()` passes that filename to `govc import.ova -options=<file>` but never removed the file afterward, leaving the password on disk after `om` exits.

`CreateVM()` now removes the temp options file via `defer` immediately after it's created, so it's cleaned up regardless of how the rest of `CreateVM()` returns (success, error, or early exit).

Two follow-up commits address review feedback on this PR:

- **Restrict vSphere options temp file permissions to 0600** (`e532c531`): `createOptionsFile()` wrote the options JSON with `os.WriteFile(..., 0644)`. In practice the file was already `0600` because `os.CreateTemp` creates it first and `WriteFile`'s mode argument only applies on file creation, not when writing to an existing file — but the `0644` argument was misleading. Changed it to `0600` to make the intended permission explicit and added a regression test asserting the file stays owner-only readable while it exists.
- **Clean up temp vSphere CA cert file after govc calls** (`1b3c387b`): `addEnvVars()` wrote the vCenter CA cert to a temp file (for `GOVC_TLS_CA_CERTS`) but never removed it, leaking a temp file on every `CreateVM`/`DeleteVM` call. `addEnvVars()` now also returns that file's path, and both callers defer its removal after all `govc` calls complete, mirroring the options-file cleanup pattern. Added a regression test asserting the CA cert temp file is removed after `CreateVM()`.

## Test plan

- [x] Added a test asserting the temp options file no longer exists after `CreateVM()` returns — confirmed it fails without the fix and passes with it
- [x] Added a test asserting the temp options file is written with `0600` permissions — confirmed green
- [x] Added a test asserting the temp CA cert file no longer exists after `CreateVM()` returns — confirmed it fails without the fix and passes with it
- [x] `go vet ./vmlifecycle/vmmanagers/...` — clean
- [x] `go test ./vmlifecycle/vmmanagers/...` — 229/232 passing; the remaining 3 failures (aws/azure/openstack config-selection tests) are pre-existing and unrelated to this change — confirmed identical failures before these commits

ai-assisted=yes

### DIFF
--- a/vmlifecycle/vmmanagers/vsphere.go
+++ b/vmlifecycle/vmmanagers/vsphere.go
@@ -166,6 +166,11 @@ func (v *VsphereVMManager) CreateVM() (Status, StateInfo, error) {
 	if err != nil {
 		return Unknown, StateInfo{}, err
 	}
+	defer func() {
+		if removeErr := os.Remove(optionFilename); removeErr != nil {
+			log.Printf("could not remove temp options file %s: %s", optionFilename, removeErr)
+		}
+	}()
 
 	ipath := v.createIpath()
 
--- a/vmlifecycle/vmmanagers/vsphere_test.go
+++ b/vmlifecycle/vmmanagers/vsphere_test.go
@@ -7,6 +7,7 @@ import (
 	"fmt"
 	"io"
 	"os"
+	"strings"
 
 	. "github.com/onsi/ginkgo/v2"
 	. "github.com/onsi/gomega"
@@ -72,6 +73,24 @@ opsman-configuration:
 					command, _ := createCommand(configStr, opsmanVersionBelow26)
 					_, _, err := command.CreateVM()
 					Expect(err).ToNot(HaveOccurred())
+				})
+
+				It("removes the temp options file (which contains the cleartext admin_password) after use", func() {
+					command, runner := createCommand(configStr, opsmanVersionBelow26)
+					_, _, err := command.CreateVM()
+					Expect(err).ToNot(HaveOccurred())
+
+					_, args := runner.ExecuteWithEnvVarsArgsForCall(0)
+					var optionsFilename string
+					for _, arg := range args {
+						if s, ok := arg.(string); ok && strings.HasPrefix(s, "-options=") {
+							optionsFilename = strings.TrimPrefix(s, "-options=")
+						}
+					}
+					Expect(optionsFilename).ToNot(BeEmpty())
+
+					_, statErr := os.Stat(optionsFilename)
+					Expect(os.IsNotExist(statErr)).To(BeTrue(), "expected temp options file to be removed after CreateVM")
 				})
 
 				It("calls govc with correct cli arguments, and does not duplicate /datacenter/vm path", func() {


vmlifecycle/vmmanagers/vsphere.go
```diff
@@ -166,6 +166,11 @@ func (v *VsphereVMManager) CreateVM() (Status, StateInfo, error) {
 	if err != nil {
 		return Unknown, StateInfo{}, err
 	}
+	defer func() {
+		if removeErr := os.Remove(optionFilename); removeErr != nil {
```
skaarthik: 🤖 AI Review
---
🟡 **Should Fix**

Good fix for the disclosure window this PR targets, but note there's a related exposure that isn't addressed here: `createOptionsFile` (line ~261) writes the options JSON — which embeds this same cleartext `admin_password` for legacy `ssh_password` deployments — with `os.WriteFile(optionsFile.Name(), optionsBytes, 0644)`. `0644` is world-readable, so on a multi-user host any local user can read the password during the (now shortened, but nonzero) window between file creation and this `defer os.Remove`. Since this PR is specifically about closing an info-disclosure vulnerability for this exact value, consider tightening that mode to `0600` too — small change, same code path, closes a related exposure.

skaarthik: 🤖 AI Review
---
🟢 **Suggestion**

`addEnvVars` (line ~298) also creates a temp file (`ca.crt`) via `os.CreateTemp` that is never removed — same leaked-temp-file pattern being fixed here, just for a CA certificate instead of a password (lower sensitivity, but still worth a `defer os.Remove(caCertFile.Name())` for consistency/disk hygiene). Not blocking since it's outside this PR's stated scope (TNZGOV-8041), just flagging since it's the same bug class in the same file.

skaarthik: 🤖 AI Review
---
🟢 **Suggestion**

Good placement — the `defer` is registered immediately after `createOptionsFile()` succeeds, so it covers every return path below (including the early `Exist`/`Incomplete` returns) without any defer-before-file-exists ordering risk. No changes needed, just confirming this covers all the paths I traced.

vmlifecycle/vmmanagers/vsphere.go
```diff
@@ -257,7 +276,7 @@ func (v *VsphereVMManager) createOptionsFile() (optionsFileName string, err erro
 		return "", fmt.Errorf("could not create temp option file: %s", err)
 	}
 
-	err = os.WriteFile(optionsFile.Name(), optionsBytes, 0644)
+	err = os.WriteFile(optionsFile.Name(), optionsBytes, 0600)
```
krsri, line 279: 🤖 AI Review
---
🟡 **Should Fix**

`os.CreateTemp` already creates and opens `optionsFile` with mode `0600`. Calling `os.WriteFile(optionsFile.Name(), optionsBytes, 0600)` re-opens the exact same file path while the original file handle remains open.

Additionally, if `os.WriteFile` fails, `optionsFile.Close()` is skipped, leaving both the file descriptor leaked and the temporary file orphaned on disk (since `createOptionsFile()` returns `("", err)` and `CreateVM()` never reaches its `defer os.Remove`). Also, the error message on line 288 (`optionsFile.Close()`) currently says `"could not write options to file"` instead of `"could not close options file"`.

Consider writing directly to `optionsFile`, closing it, and cleaning up on error:

```go
	optionsFile, err := os.CreateTemp("", "options.json")
	if err != nil {
		return "", fmt.Errorf("could not create temp option file: %s", err)
	}
	optionsFileName = optionsFile.Name()

	if _, err = optionsFile.Write(optionsBytes); err != nil {
		_ = optionsFile.Close()
		_ = os.Remove(optionsFileName)
		return "", fmt.Errorf("could not write options to file: %s", err)
	}

	if err = optionsFile.Close(); err != nil {
		_ = os.Remove(optionsFileName)
		return "", fmt.Errorf("could not close options file: %s", err)
	}

	return optionsFileName, nil
```

vmlifecycle/vmmanagers/vsphere.go
```diff
@@ -292,16 +311,17 @@ func (v *VsphereVMManager) addEnvVars() (envVarsList []string, err error) {
 	if v.Config.OpsmanConfig.Vsphere.Vcenter.CACert != "" {
 		caCertFile, err := os.CreateTemp("", "ca.crt")
 		if err != nil {
-			return []string{}, fmt.Errorf("could not create temp file for ca cert: %s", err)
+			return []string{}, "", fmt.Errorf("could not create temp file for ca cert: %s", err)
 		}
 		_, err = caCertFile.WriteString(v.Config.OpsmanConfig.Vsphere.Vcenter.CACert)
```
krsri, line 316: 🤖 AI Review
---
🔴 **Must Fix**

`caCertFile` is created via `os.CreateTemp("", "ca.crt")` but is never closed with `caCertFile.Close()`.

This causes two issues:
1. **Windows Deletion Failure:** `om` builds Windows binaries (`om-windows-amd64`). On Windows, deleting an open file via `os.Remove()` fails with a sharing violation (`ERROR_SHARING_VIOLATION` / `Access is denied`). The `defer os.Remove(caCertFilename)` in `CreateVM()` and `DeleteVM()` will fail on Windows while the handle is open.
2. **Orphaned File on Write Failure:** If `caCertFile.WriteString(...)` fails, `addEnvVars()` returns `caCertFilename = ""`. The caller never receives the path to delete it, and the file descriptor is leaked.

Consider closing the file descriptor after writing and ensuring cleanup on error:

```go
	if v.Config.OpsmanConfig.Vsphere.Vcenter.CACert != "" {
		caCertFile, err := os.CreateTemp("", "ca.crt")
		if err != nil {
			return []string{}, "", fmt.Errorf("could not create temp file for ca cert: %s", err)
		}
		caCertFilename = caCertFile.Name()

		if _, err = caCertFile.WriteString(v.Config.OpsmanConfig.Vsphere.Vcenter.CACert); err != nil {
			_ = caCertFile.Close()
			_ = os.Remove(caCertFilename)
			return []string{}, "", fmt.Errorf("could not write cert to the cert file: %s", err)
		}

		if err = caCertFile.Close(); err != nil {
			_ = os.Remove(caCertFilename)
			return []string{}, "", fmt.Errorf("could not close ca cert file: %s", err)
		}

		env = append(env, "GOVC_TLS_CA_CERTS="+caCertFilename)
	}
```

vmlifecycle/vmmanagers/vsphere.go
skaarthik: Fixed in 539c6b92 — createOptionsFile() now writes directly to the open temp-file handle (instead of reopening the same path via os.WriteFile), closes it, and removes the temp file on any write/close error. Also fixed the copy-pasted error message on the Close() failure path (now "could not close options file").

vmlifecycle/vmmanagers/vsphere.go
skaarthik: Fixed in 539c6b92 — addEnvVars() now closes caCertFile after writing to it (avoids the Windows os.Remove-on-open-handle sharing-violation issue) and removes the temp file if either the write or the close fails, so it no longer gets orphaned on error.

krsri (approved):

PR #825, opened 2026-09-09
fix(TNZGOV-4038): reject path traversal via tile metadata in config-template

skaarthik:
## Summary

Fixes TNZGOV-4038 — Path traversal via tile metadata in config-template output (path-traversal, LOW/P2).

`Executor.Generate()` in `configtemplate/generator/execute.go` took `productName`/`productVersion` straight from a `.pivotal` tile's metadata YAML and used them as filesystem path segments via `path.Join(e.baseDirectory, productName, productVersion)`, then `os.MkdirAll`/`os.WriteFile`. `path.Join` cleans the result but does not prevent `..` from escaping the base directory. The same issue applied to the map keys used as ops-file names when writing network/resource/features/optional YAML files.

A crafted tile with `name: "../../../../home/ci/.config"` could cause `om config-template` to create directories and write YAML files outside the operator-specified `--output-directory`.

## Fix

Added `validateMetadataPathSegment()`, which rejects any tile-metadata-derived value used as a path segment if it contains `..` or a path separator. Applied it to:
- `productName` / `productVersion` before they're joined into `targetDirectory`
- each `name` key from `networkOpsFiles`, `resourceOpsFiles`, `productPropertyOpsFiles`, and `productPropertyOptionalOpsFiles` before being used as a filename

This closes every traversal vector identified in the finding (the product/version join plus all four ops-file write sites).

## BREAKING CHANGES

**This PR introduces a breaking change to `om config-template`.**

- **Before:** `om config-template` would accept any tile metadata, including a product `name`/`product_version`, job name, or property/option name containing `..`, `/`, `\`, or `:`. Generation always succeeded — in the traversal case, it silently wrote files outside the intended `--output-directory`; in the benign-looking `/`/`:` case, it wrote into an unexpected nested subdirectory rather than the flat ops-file layout.
- **After:** `om config-template` now returns an error ("path traversal detected in tile metadata: invalid ...") and generates nothing for any tile whose metadata contains one of these characters in a field used to build an output path or filename.
- **Who is affected:** Anyone running `om config-template` against a `.pivotal` tile whose metadata legitimately (not maliciously) uses `..`, `/`, `\`, or `:` in the product name, product version, a job name, or a property/multi-select option name. This is expected to be rare in practice (product versions are typically semver-like, e.g. `1.2.3-build.4`), but any tile author or CI pipeline relying on such a value will see a new failure where generation previously succeeded.
- **What to do:** If `om config-template` starts failing with a "path traversal detected in tile metadata" error after upgrading, inspect the referenced field in the tile's `metadata.yml` (or the tile author's job/property names) and remove the offending character. There is no flag to opt out of this validation — the fix is intentionally unconditional because the rejected characters are exactly the ones capable of causing incorrect or unsafe file writes.

Two follow-up commits address review feedback on this PR:

- **Add regression tests for ops-file path traversal via job/property names** (`87437f1d`): the original test only covered the `productName`/`productVersion` fields. Added coverage for the other four guarded sites — a job name containing `/` flowing into a `resourceOpsFiles` key (e.g. `evil/job_elb_names`), and a multi-select property option name flowing into a `productPropertyOpsFiles` key. Both are confirmed to fail without the guard (attempting to write into an unintended nested directory) and pass with it.
- **Reject colons in tile-metadata path segments** (`40213ea8`): `validateMetadataPathSegment` allowed values containing `:`, which on Windows/NTFS can reference an Alternate Data Stream (e.g. `filename:stream`). Rejected it alongside `..`, `/`, and `\`, with a regression test using a colon-containing product name.

## Test plan

- [x] Added a test with a malicious `name`/`product_version` (`../escaped-product` / `../escaped-version`) asserting `Generate()` returns a "path traversal" error and that nothing is written outside the base directory — confirmed it fails without the fix and passes with it
- [x] Added a test asserting a job name containing `/` is rejected before it can be used as a `resourceOpsFiles` filename — confirmed it fails without the guard and passes with it
- [x] Added a test asserting a multi-select property option name containing `/` is rejected before it can be used as a `productPropertyOpsFiles` filename — confirmed it fails without the guard and passes with it
- [x] Added a test asserting a colon in the product name is rejected (Windows ADS protection) — confirmed it fails without the check and passes with it
- [x] `go vet ./configtemplate/...` — clean
- [x] `gofmt -l` on changed files — no output
- [x] `go test ./configtemplate/generator/...` — 96/96 passing; stable across 5 repeated runs (no map-iteration-order flakiness); no regressions

ai-assisted=yes

### DIFF
--- a/configtemplate/generator/execute.go
+++ b/configtemplate/generator/execute.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 	"os"
 	"path"
+	"strings"
 
 	"gopkg.in/yaml.v2"
 )
@@ -38,8 +39,14 @@ func (e *Executor) Generate() error {
 	if productVersion == "" {
 		return errors.New("version in metadata is blank")
 	}
+	if err := validateMetadataPathSegment("product version", productVersion); err != nil {
+		return err
+	}
 
 	productName := metadata.ProductName()
+	if err := validateMetadataPathSegment("product name", productName); err != nil {
+		return err
+	}
 
 	targetDirectory := path.Join(e.baseDirectory, productName)
 	if !e.doNotIncludeProductVersion {
@@ -86,6 +93,9 @@ func (e *Executor) Generate() error {
 
 	if len(networkOpsFiles) > 0 {
 		for name, contents := range networkOpsFiles {
+			if err = validateMetadataPathSegment("network ops-file name", name); err != nil {
+				return err
+			}
 			if err = e.writeYamlFile(path.Join(networkDirectory, fmt.Sprintf("%s.yml", name)), contents); err != nil {
 				return err
 			}
@@ -115,6 +125,9 @@ func (e *Executor) Generate() error {
 
 	if len(resourceOpsFiles) > 0 {
 		for name, contents := range resourceOpsFiles {
+			if err = validateMetadataPathSegment("resource ops-file name", name); err != nil {
+				return err
+			}
 			if err = e.writeYamlFile(path.Join(resourceDirectory, fmt.Sprintf("%s.yml", name)), contents); err != nil {
 				return err
 			}
@@ -145,6 +158,9 @@ func (e *Executor) Generate() error {
 
 	if len(productPropertyOpsFiles) > 0 {
 		for name, contents := range productPropertyOpsFiles {
+			if err = validateMetadataPathSegment("product property ops-file name", name); err != nil {
+				return err
+			}
 			if err = e.writeYamlFile(path.Join(featuresDirectory, fmt.Sprintf("%s.yml", name)), contents); err != nil {
 				return err
 			}
@@ -158,6 +174,9 @@ func (e *Executor) Generate() error {
 
 	if len(productPropertyOptionalOpsFiles) > 0 {
 		for name, contents := range productPropertyOptionalOpsFiles {
+			if err = validateMetadataPathSegment("product property optional ops-file name", name); err != nil {
+				return err
+			}
 			if err = e.writeYamlFile(path.Join(optionalDirectory, fmt.Sprintf("%s.yml", name)), contents); err != nil {
 				return err
 			}
@@ -187,6 +206,17 @@ func (e *Executor) CreateTemplate(metadata *Metadata) (*Template, error) {
 		template.ErrandConfig = CreateErrandConfig(metadata)
 	}
 	return template, nil
+}
+
+// validateMetadataPathSegment rejects tile-metadata-derived values that are
+// used as filesystem path segments (product name/version, ops-file names) if
+// they could escape the intended output directory via ".." or a path
+// separator.
+func validateMetadataPathSegment(fieldName, value string) error {
+	if strings.Contains(value, "..") || strings.ContainsRune(value, '/') || strings.ContainsRune(value, '\\') {
+		return fmt.Errorf("path traversal detected in tile metadata: invalid %s %q", fieldName, value)
+	}
+	return nil
 }
 
 func (e *Executor) createDirectory(path string) error {
--- a/configtemplate/generator/execute_test.go
+++ b/configtemplate/generator/execute_test.go
@@ -96,6 +96,21 @@ var _ = Describe("Executor", func() {
 			}
 		})
 
+		It("rejects tile metadata that attempts path traversal via product name or version", func() {
+			maliciousMetadata := []byte(`
+name: ../escaped-product
+product_version: ../escaped-version
+`)
+			gen := generator.NewExecutor(maliciousMetadata, tmpPath, false, true, 10, false)
+			err := gen.Generate()
+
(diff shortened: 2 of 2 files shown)


configtemplate/generator/execute_test.go
```diff
@@ -96,6 +96,21 @@ var _ = Describe("Executor", func() {
 			}
 		})
 
+		It("rejects tile metadata that attempts path traversal via product name or version", func() {
```
skaarthik, line 99: 🤖 AI Review
---
🟡 **Should Fix**

Only the product-name/version traversal path is covered by a test. The PR description says the same `validateMetadataPathSegment` guard was also applied to the four ops-file write sites (`networkOpsFiles`, `resourceOpsFiles`, `productPropertyOpsFiles`, `productPropertyOptionalOpsFiles` in execute.go), but there's no regression test proving a malicious job/property-derived ops-file name (e.g. a tile job with `name: "../evil"`, which flows into `AddResourceOpsFiles`'s `%s_elb_names` key) is actually rejected. Since this is the same vulnerability class the PR is fixing, and those code paths derive the key from attacker-controlled tile metadata via `fmt.Sprintf`, it'd be worth adding at least one test exercising one of those four sites (a crafted `job_types[].name` or property reference containing `..` or `/`) to close the coverage gap and guard against future refactors accidentally removing the check there.

configtemplate/generator/execute.go
```diff
@@ -189,6 +208,17 @@ func (e *Executor) CreateTemplate(metadata *Metadata) (*Template, error) {
 	return template, nil
 }
 
+// validateMetadataPathSegment rejects tile-metadata-derived values that are
+// used as filesystem path segments (product name/version, ops-file names) if
+// they could escape the intended output directory via ".." or a path
+// separator.
+func validateMetadataPathSegment(fieldName, value string) error {
+	if strings.Contains(value, "..") || strings.ContainsRune(value, '/') || strings.ContainsRune(value, '\\') {
```
skaarthik, line 216: 🤖 AI Review
---
🟢 **Suggestion**

Good approach validating the final derived string (map key) rather than the raw metadata field — this means it also catches traversal introduced via the `fmt.Sprintf("%s_elb_names", jobName)`-style key construction in `resource_config.go`/`product_properties.go`, not just the two direct fields. No change needed, just noting why this covers more ground than it might look like at first glance.

configtemplate/generator/execute_test.go
skaarthik: 🤖 AI Review
---
🟢 **Suggestion**

While the test coverage for path traversal via product name and version is excellent, it would be great to also add test cases for path traversal via:
1. Malicious job names (which affect `resourceOpsFiles` names).
2. Malicious property names (which affect `productPropertyOpsFiles` and `productPropertyOptionalOpsFiles` names).

This would ensure that the validation of all ops-file names is fully covered by the test suite.

configtemplate/generator/execute.go
```diff
@@ -189,6 +208,17 @@ func (e *Executor) CreateTemplate(metadata *Metadata) (*Template, error) {
 	return template, nil
 }
 
+// validateMetadataPathSegment rejects tile-metadata-derived values that are
+// used as filesystem path segments (product name/version, ops-file names) if
+// they could escape the intended output directory via ".." or a path
+// separator.
+func validateMetadataPathSegment(fieldName, value string) error {
```
skaarthik: 🤖 AI Review
---
🟢 **Suggestion**

To provide extra robustness for operators running `om` on Windows, consider also rejecting colons (`:`) in `validateMetadataPathSegment`.

On Windows/NTFS, colons are used to specify Alternate Data Streams (ADS) (e.g., `filename:stream`). Rejecting colons prevents any potential ADS manipulation or unexpected behavior on Windows filesystems.

krsri (approved):

configtemplate/generator/execute.go
```diff
@@ -189,6 +208,17 @@ func (e *Executor) CreateTemplate(metadata *Metadata) (*Template, error) {
 	return template, nil
 }
 
+// validateMetadataPathSegment rejects tile-metadata-derived values that are
+// used as filesystem path segments (product name/version, ops-file names) if
+// they could escape the intended output directory via ".." or a path
+// separator, or manipulate a Windows NTFS Alternate Data Stream via ":".
+func validateMetadataPathSegment(fieldName, value string) error {
```
yharish991-brcm: nit: fieldName is a bit misleading here, it's not an actual field from the tile metadata schema, just a human readable label we use to build the error message. Might be worth renaming to label to avoid confusion for whoever reads this next.

skaarthik: Good catch — renamed `fieldName` to `label` in `validateMetadataPathSegment` since it's just a human-readable description used to build the error message, not an actual tile metadata field. Fixed in 179dab11.

PR #830, opened 2026-09-14
TNZGOV-17460 Shell-quote bosh-env output to prevent command injection

skaarthik:
## Summary
- `om bosh-env` is documented to be consumed via `eval "$(om bosh-env)"` / `iex $(om bosh-env | Out-String)`. The values it prints (`BOSH_CLIENT`, `BOSH_CLIENT_SECRET`, `BOSH_ENVIRONMENT`, `CREDHUB_*`) come verbatim from the Ops Manager API response and were rendered with **no shell escaping** in `renderers/posix.go` and `renderers/powershell.go`.
- A malicious or MITM'd Ops Manager response (e.g. when `--skip-ssl-validation` is used) containing shell metacharacters (`;`, `$()`, backticks, `"`) in one of these values would be executed with the operator's/CI's privileges when the output is `eval`'d — CWE-78 command injection.
- Fix: both renderers now always wrap the emitted value in single quotes, escaping any embedded single quote (`'` → `'\''` for posix, `'` → `''` for powershell), instead of only quoting multi-line values. This closes the injection vector for the (previously unquoted) single-line case while keeping the existing multi-line quoting behavior.
- `commands/bosh_environment.go` needed no change — the fix is fully contained in the renderers, per the ticket's own recommendation.

Fixes TNZGOV-17460.

## BREAKING CHANGES

**This PR introduces a breaking change to the text output of `om bosh-env`.**

- **Before:** Single-line environment variable values were printed unquoted, e.g. `export BOSH_CLIENT=abc` (posix) or `$env:BOSH_CLIENT=abc` (powershell). Only multi-line values were wrapped in quotes.
- **After:** Every value, single-line or multi-line, is now wrapped in single quotes: `export BOSH_CLIENT='abc'` / `$env:BOSH_CLIENT='abc'`.
- **Who is affected:** This is safe, and produces the identical resulting environment variable value, for the documented consumption pattern (`eval "$(om bosh-env)"` / `iex $(om bosh-env | Out-String)`) — shell quoting doesn't change what gets assigned. It **does** affect any consumer that parses the raw text output directly instead of evaluating it — e.g. a script doing `om bosh-env | grep BOSH_CLIENT= | cut -d= -f2` or a regex like `BOSH_CLIENT=(\S+)` will now capture `'abc'` (including the quotes) instead of `abc`.
- **What to do:** Consumers should evaluate the output as documented (`eval`/`iex`) rather than scrape it as text. If a script must parse the raw text, it should strip a single leading and trailing `'` from the captured value and unescape any `''`/`'\''` sequences back to a literal `'`, rather than treating the substring after `=` as the literal value.

One follow-up commit addresses review feedback on this PR:

- **Fix extra blank line in powershell multi-line trailing newline** (`62fcc567`): `RenderEnvironmentVariable` in `renderers/powershell.go` normalized a trailing newline to `\r\n` by checking `!strings.HasSuffix(value, "\r\n")`, which is also `true` when `value` already ends in a bare `\n` — appending a second `\r\n` and producing an extra blank line at the end of multi-line values. Now checks for a bare `\n` first and only replaces it, leaving already-correct `\r\n` endings untouched.

## Test plan
- [x] Added failing tests in `renderers/posix_test.go` / `renderers/powershell_test.go` covering shell-metacharacter payloads and embedded single quotes; confirmed red against unpatched code.
- [x] Implemented the fix; `go test ./renderers/... ./commands/...` passes (620 specs, including updated `commands/bosh_environment_test.go` expectations for the new quoted output format).
- [x] Manual PoC: rendered a `BOSH_CLIENT_SECRET` value of `x;touch /tmp/pwned-om-poc` through the posix renderer and `eval`'d the resulting line in bash — confirmed the injected command no longer executes.
- [x] Added a test asserting a multi-line value ending in a bare `\n` is normalized to a single trailing `\r\n` (no extra blank line) in the powershell renderer — confirmed it fails without the fix and passes with it
- [x] `go vet ./renderers/...` — clean
- [x] `gofmt -l` on changed files — no output

ai-assisted=yes


renderers/powershell.go
```diff
@@ -15,13 +15,19 @@ func NewPowershell() Renderer {
 
 func (renderer *powershell) RenderEnvironmentVariable(variable string, value string) string {
 	if strings.ContainsAny(value, "\n") {
-		suffix := ""
-		if !strings.HasSuffix(value, "\n") {
-			suffix = "\r\n"
+		if !strings.HasSuffix(value, "\r\n") {
+			value += "\r\n"
```
skaarthik: 🤖 AI Review
---
🟡 **Should Fix**

If `value` ends with a Unix newline (`\n`) but not a Windows newline (`\r\n`), `!strings.HasSuffix(value, "\r\n")` evaluates to `true`, which appends `\r\n` to the end of `value`. This results in a trailing `\n\r\n` (an extra empty line at the end of the value).

To normalize the trailing newline to `\r\n` without adding an extra empty line, consider replacing the trailing `\n` with `\r\n` when it is present:

```go
	if strings.ContainsAny(value, "\n") {
		if !strings.HasSuffix(value, "\n") {
			value += "\r\n"
		} else if !strings.HasSuffix(value, "\r\n") {
			value = value[:len(value)-1] + "\r\n"
		}
		value = "\r\n" + value
	}
```

krsri (approved):

yharish991-brcm (approved):

PR #837, opened 2026-09-18
Add --fix-stemcells flag to apply-changes

Alphasite:
## The failure this addresses

If a stemcell is not usable we need a way to force it to be re-uploaded. `bosh upload-stemcell --fix` is the
existing escape hatch, but Apply Changes drives the uploads, so there was no way to ask for it from `om`.

Ops Manager 11.0 added an optional boolean `fix_stemcells` to `POST /api/v0/installations`: when true, every
stemcell upload in that run uses `--fix`. It does not persist — it applies to exactly one Apply Changes run,
unlike the director properties `--recreate-vms` writes.

## This change

`om apply-changes --fix-stemcells` sets it, following the `--allow-unsafe-dependency-*` precedent end to end:
a bool on `ApplyChanges.Options`, a new positional parameter on `CreateInstallation`, and a
`fix_stemcells,omitempty` request-body field, so an unset flag is absent from the payload rather than an
explicit `false`. The flag name is plural, matching the API property rather than bosh's singular `--fix`.

It is gated on Ops Manager 11.0, the first release to support the property. An older target would silently
ignore it and leave the operator believing the stemcell was re-uploaded. The `Info()` call happens only when
the flag is set, so an ordinary `apply-changes` is unaffected. The gate runs before the reattach path, so
`--fix-stemcells --reattach` errors below 11.0; at 11.0+ reattach returns before `CreateInstallation`, so the
flag has no effect there, matching the sibling flags. A log line is emitted when the flag is active, as those
flags do.

Tests cover flag propagation, the log line, the sub-11.0 error, and `fix_stemcells` presence/absence in the
POST body. The counterfeiter fake for `applyChangesService` is regenerated. `go test ./api/... ./commands/...`
passes.


aramprice (approved):
seems reasonable

davewalter (approved):

commands/apply_changes.go
```diff
@@ -28,13 +28,14 @@ type ApplyChanges struct {
 		ForceLatestVariables          bool     `long:"force-latest-variables" description:"force any certificates or other BOSH variables to use their latest version even when a stemcell is not being upgraded"`
 		AllowUnsafeDependencyUpdate   bool     `long:"allow-unsafe-dependency-update" description:"allow apply-changes to proceed even when it would update a dependency out of its declared safe order"`
 		AllowUnsafeDependencyDeletion bool     `long:"allow-unsafe-dependency-deletion" description:"allow apply-changes to proceed even when it would delete an optional dependency not marked as safe to delete"`
+		FixStemcells                  bool     `long:"fix-stemcells" description:"upload stemcells with --fix, forcing the director to re-upload them even when it already has the same name and version"`
```
drich10: 🤖 AI Review
---
🟡 **Should Fix**

Consider adding a version gate for Ops Manager 11.0+ (`info.VersionAtLeast(11, 0)`), matching `--allow-unsafe-dependency-*` and `--product-name`.

The PR description mentions that no `VersionAtLeast` gate was added because *"an older Ops Manager simply ignores the unknown property, and gating would add an Info() round trip to every invocation."* However:

1. **No overhead on normal runs:** Gating only checks `ac.service.Info()` when `ac.Options.FixStemcells` is true (inside an `if` block, just like `--allow-unsafe-dependency-*`), adding zero overhead to standard `apply-changes` runs.
2. **Prevent silent failures:** When targeting an Ops Manager older than 11.0, Ops Manager will silently ignore the unknown `fix_stemcells` JSON field and perform a normal apply-changes without `--fix`. An operator running this flag (e.g. to repair stemcells or unblock a cross-vCenter AZ migration per TNZ-147487) will believe the fix was applied, while the underlying stemcell was never re-uploaded to the target vCenter.
3. **Dead test stubbing:** In `commands/apply_changes_test.go` (lines 176 and 190), `service.InfoReturns(api.Info{Version: "11.0"}, nil)` is already stubbed from the copied test template, but is currently unused because the check is missing.

api/installations_service.go
```diff
@@ -112,7 +112,7 @@ func (a Api) ListInstallations() ([]InstallationsServiceOutput, error) {
 	return responseStruct.Installations, nil
 }
 
-func (a Api) CreateInstallation(ignoreWarnings bool, deployProducts bool, forceLatestVariables bool, allowUnsafeDependencyUpdate bool, allowUnsafeDependencyDeletion bool, productNames []string, errands ApplyErrandChanges) (InstallationsServiceOutput, error) {
+func (a Api) CreateInstallation(ignoreWarnings bool, deployProducts bool, forceLatestVariables bool, allowUnsafeDependencyUpdate bool, allowUnsafeDependencyDeletion bool, fixStemcells bool, productNames []string, errands ApplyErrandChanges) (InstallationsServiceOutput, error) {
```
drich10: 🤖 AI Review
---
🟡 **Should Fix**

`CreateInstallation` now takes 6 consecutive `bool` arguments (`ignoreWarnings`, `deployProducts`, `forceLatestVariables`, `allowUnsafeDependencyUpdate`, `allowUnsafeDependencyDeletion`, `fixStemcells`).

This makes calls like `CreateInstallation(false, true, false, false, false, false, nil, errands)` error-prone to argument transposition and makes tests and fakes brittle to maintain. While this follows the recent precedent from #822, strongly consider grouping installation settings into an options struct (e.g. `CreateInstallationOptions` or `InstallationParams`) to improve type safety, readability, and future maintainability.

commands/apply_changes.go
```diff
@@ -181,7 +182,7 @@ func (ac ApplyChanges) Execute(args []string) error {
 	}
 
 	ac.logger.Printf("attempting to apply changes to the targeted Ops Manager")
-	installation, err = ac.service.CreateInstallation(ac.Options.IgnoreWarnings, !ac.Options.SkipDeployProducts, ac.Options.ForceLatestVariables, ac.Options.AllowUnsafeDependencyUpdate, ac.Options.AllowUnsafeDependencyDeletion, changedProducts, errands)
+	installation, err = ac.service.CreateInstallation(ac.Options.IgnoreWarnings, !ac.Options.SkipDeployProducts, ac.Options.ForceLatestVariables, ac.Options.AllowUnsafeDependencyUpdate, ac.Options.AllowUnsafeDependencyDeletion, ac.Options.FixStemcells, changedProducts, errands)
```
drich10, line 202: 🤖 AI Review
---
🟢 **Suggestion**

Consider logging when `--fix-stemcells` is active before triggering the installation (similar to lines 177-182 for `allow-unsafe-dependency-*`):

```go
if ac.Options.FixStemcells {
	ac.logger.Printf("fix-stemcells is set: stemcells will be uploaded with --fix")
}
```

Since stemcell uploads can significantly increase the duration of an Apply Changes run, emitting a log line confirms to the operator (especially in CI/CD pipeline logs) that the flag was acknowledged and stemcells are being re-uploaded.

docs/apply-changes/README.md
```diff
@@ -89,6 +89,10 @@ Help Options:
                                               even when it would delete an
                                               optional dependency not marked as
                                               safe to delete
+          --fix-stemcells                     upload stemcells with --fix,
```
drich10, line 92: 🤖 AI Review
---
🟢 **Suggestion**

1. Other version-dependent flags in the documentation and CLI help text note the minimum Ops Manager version (e.g. `--product-name ... (OM 2.2+)`). Consider annotating `--fix-stemcells` with `(OM 11.0+)` here and in `commands/apply_changes.go` description.
2. Consider adding an entry under `## Unreleased -> ### Features` in `CHANGELOG.md` to document the new `--fix-stemcells` flag.

commands/apply_changes.go
Alphasite: Agreed, and fixed in 0128085 — the overhead argument in the description was wrong, since the `Info()` call sits inside `if ac.Options.FixStemcells` and so costs nothing on a normal run.

I checked the support boundary rather than assuming it by analogy: `fix_stemcells` first shipped in Ops Manager 11.0 and was not backported to any earlier line, so `VersionAtLeast(11, 0)` is the real boundary and the gate is safe. It follows the two-branch shape of the `--allow-unsafe-dependency-*` gate, wrapping with `%w` only when the version is undeterminable, so the version-too-old message carries no trailing `Error: %!w(<nil>)`. The previously-dead `InfoReturns` stubbing is now live, and there is a new test asserting the exact error against a 10.2 target with `CreateInstallation` never called.

commands/apply_changes.go
Alphasite: Done in 0128085: `fix-stemcells is set: stemcells will be uploaded with --fix`, emitted alongside the `allow-unsafe-dependency-*` lines and asserted in the flag test.

docs/apply-changes/README.md
Alphasite: Both done in 0128085. The flag description now ends `(OM 11.0+)`, in `commands/apply_changes.go` and in this help paste — I regenerated `--help` to confirm the added text does not change the column wrapping. A CHANGELOG entry is added under `## Unreleased` -> `### Features`, noting that the flag applies to a single Apply Changes run and requires 11.0 or later.

Alphasite:
Squashed to a single commit `3b0ffe7`, with a message describing the final state (including the 11.0 gate) rather than the superseded no-gate reasoning. My review replies above cite `0128085`; that work is now in `3b0ffe7`. Also dropped a Vim swap file that had been accidentally staged. Same tree, no source changes since the last push.

commands/apply_changes.go
```diff
@@ -28,13 +28,14 @@ type ApplyChanges struct {
 		ForceLatestVariables          bool     `long:"force-latest-variables" description:"force any certificates or other BOSH variables to use their latest version even when a stemcell is not being upgraded"`
 		AllowUnsafeDependencyUpdate   bool     `long:"allow-unsafe-dependency-update" description:"allow apply-changes to proceed even when it would update a dependency out of its declared safe order"`
 		AllowUnsafeDependencyDeletion bool     `long:"allow-unsafe-dependency-deletion" description:"allow apply-changes to proceed even when it would delete an optional dependency not marked as safe to delete"`
+		FixStemcells                  bool     `long:"fix-stemcells" description:"upload stemcells with --fix, forcing the director to re-upload them even when it already has the same name and version (OM 11.0+)"`
```
drich10: 🤖 AI Review
---
🔴 **Must Fix**

An accidental Vim swap file `commands/.apply_changes.go.swp` (16 KB) was committed in `259fceee`.

Please delete `commands/.apply_changes.go.swp` from the branch before merging (and consider adding `*.swp` to `.gitignore`).

drich10 (approved):

